PR #16, opened 2020-06-16
Improve calculations

D4nte:
Improve mathematics by avoiding floating point calculations and state expectations around precision.

TODO: property testing.

### DIFF
--- a/src/bitcoin.rs
+++ b/src/bitcoin.rs
@@ -1,0 +1,99 @@
+use crate::dai;
+use crate::dai::ATTOS_IN_DAI_EXP;
+use crate::float_maths::multiple_pow_ten;
+use crate::publish::WorthIn;
+use num::pow::Pow;
+use num::BigUint;
+
+pub const SATS_IN_BITCOIN_EXP: u16 = 8;
+
+#[derive(Debug, Copy, Clone, Ord, PartialOrd, PartialEq, Eq)]
+pub struct Amount(::bitcoin::Amount);
+
+impl Amount {
+    pub fn from_btc(btc: f64) -> anyhow::Result<Amount> {
+        Ok(Amount(::bitcoin::Amount::from_btc(btc)?))
+    }
+
+    pub fn from_sat(sat: u64) -> Self {
+        Amount(::bitcoin::Amount::from_sat(sat))
+    }
+
+    pub fn as_sat(self) -> u64 {
+        self.0.as_sat()
+    }
+
+    pub fn as_btc(self) -> f64 {
+        self.0.as_btc()
+    }
+}
+
+impl WorthIn<dai::Amount> for Amount {
+    const MAX_PRECISION_EXP: u16 = 9;
+
+    fn worth_in(&self, btc_to_dai_rate: f64) -> anyhow::Result<dai::Amount> {
+        if btc_to_dai_rate.is_sign_negative() {
+            anyhow::bail!("Rate is negative.");
+        }
+
+        if btc_to_dai_rate <= 10e-10 {
+            anyhow::bail!("Rate is null.");
+        }
+
+        if btc_to_dai_rate.is_infinite() {
+            anyhow::bail!("Rate is infinite.");
+        }
+
+        let uint_rate =
+            multiple_pow_ten(btc_to_dai_rate, Self::MAX_PRECISION_EXP).map_err(|_| {
+                anyhow::anyhow!("Rate's precision is too high, truncation would ensue.")
+            })?;
+
+        // Apply the rate
+        let worth = uint_rate * self.as_sat();
+
+        // The rate input is for bitcoin to dai but we applied to satoshis so we need to:
+        // - divide to get bitcoins
+        // - divide to adjust for max_precision
+        // - multiple to get attodai
+        let adjustment_exp =
+            BigUint::from(ATTOS_IN_DAI_EXP - Self::MAX_PRECISION_EXP - SATS_IN_BITCOIN_EXP);
+
+        let adjustment = BigUint::from(10u64).pow(adjustment_exp);
+
+        let atto_dai = worth * adjustment;
+
+        Ok(dai::Amount::from_atto(atto_dai))
+    }
+}
+
+impl std::ops::Sub for Amount {
+    type Output = Amount;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        Amount(self.0 - rhs.0)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn using_too_precise_rate_returns_error() {
+        let btc = Amount::from_btc(1.0).unwrap();
+
+        let res: anyhow::Result<dai::Amount> = btc.worth_in(1000.1234567891);
+
+        assert!(res.is_err())
+    }
+
+    #[test]
+    fn using_rate_returns_correct_result() {
+        let btc = Amount::from_btc(1.0).unwrap();
+
+        let res: dai::Amount = btc.worth_in(1000.123456789).unwrap();
+
+        assert_eq!(res, dai::Amount::from_dai_trunc(1000.123456789).unwrap());
+    }
+}
--- a/src/dai.rs
+++ b/src/dai.rs
@@ -1,6 +1,7 @@
-use crate::float_maths::{multiple_pow_ten, truncate};
+use crate::bitcoin::{self, SATS_IN_BITCOIN_EXP};
+use crate::float_maths::{divide_pow_ten_trunc, multiple_pow_ten, truncate};
 use crate::publish::WorthIn;
-use num::{pow::Pow, BigUint};
+use num::{pow::Pow, BigUint, ToPrimitive};
 use std::ops::{Div, Mul};
 
 pub const ATTOS_IN_DAI_EXP: u16 = 18;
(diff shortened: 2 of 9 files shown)


src/bitcoin.rs
```diff
@@ -1,20 +1,36 @@
 use crate::publish::NominalMulf64;
+use crate::dai;
+
+pub const BITCOIN_DEC: u64 = 100_000_000;
 
 #[derive(Debug, Copy, Clone, Ord, PartialOrd, PartialEq, Eq)]
 pub struct Amount(u64);
```
thomaseizinger: We don't you just delegate to `rust-bitcoin` for the inner implementation? :)

src/dai.rs
```diff
@@ -1,20 +1,43 @@
...
+
+    #[allow(clippy::cast_precision_loss)] // It's ok because it just means we are applying slightly more than the given spread
+    pub fn from_dai(dai: f64) -> Self {
+        let dai_dec = BigUint::from(DAI_DEC);
+        let precision = 10u64^14;
+        let value = (dai * precision as f64).round();
+        let big_u_int = BigUint::from_f64(value).unwrap();
+
+        Amount(dai_dec.mul(big_u_int).div(precision))
+    }
```
thomaseizinger: If these functions are meant to be used in production code, I would suggest testing them with `proptest`: https://docs.rs/proptest/0.10.0/proptest/

src/publish.rs
```diff
@@ -72,7 +73,8 @@ where
     ) - bitcoin_wallet.bitcoin_fees();
 
     let rate = spread.apply(mid_market_rate);
-    let buy_amount = sell_amount.is_worth_in(rate);
+    // TODO: Return result
+    let buy_amount = sell_amount.is_worth_in(rate).unwrap();
```
thomaseizinger: Would it make sense to calculate the `buy_amount` first and then apply the spread to it?

I am asking because I remember coding on the kraken interface with @da-kami and I think the actual data we get returned is _not_ a rate but a quantity. Meaning there would be quite some back and forth where we:

- ask for a the rate from kraken, we get `1BTC = 9543 DAI`
- convert that to a rate
- apply a spread to the rate
- convert the rate back to an amount

From that perspective, it seems like we could spare ourselves quite some arithmetic if we just do:

- ask for a the rate from kraken, we get `1BTC = 9543 DAI`
- apply a spread to the amount

D4nte:
> Before I actually jump in and review the maths, I wonder if it would be possible to use something like [this](https://docs.rs/num-rational/0.3.0/num_rational/struct.Ratio.html) or [this](https://github.com/dnsl48/fraction) to avoid some of the complexity of handling floats. Did you already consider that? Would it help at all?

I considered it but the value we receive from Kraken may not be a rational fraction or ratio. Also I felt it would be faster to roll out what I did then learn the limitation of the provided libs.

src/dai.rs
```diff
@@ -8,7 +8,12 @@ pub const ATTOS_IN_DAI_EXP: u16 = 18;
 pub const DAI_DEC: u64 = 1_000_000_000_000_000_000;
 
 // It means the mantissa can be up to 9 digits long
-pub const FROM_ROUNDED_DAI_PRECISION: u16 = 9;
+const FROM_ROUNDED_DAI_PRECISION_EXP: u16 = 9;
+
+lazy_static::lazy_static! {
+    pub static ref FROM_ROUNDED_DAI_PRECISION: u32 =
+        10u32.pow(FROM_ROUNDED_DAI_PRECISION_EXP as u32);
+}
```
thomaseizinger: There is a (IMO) simpler alternative :)

https://docs.rs/conquer-once/0.2.0/conquer_once/type.Lazy.html

src/dai.rs
thomaseizinger: To elaborate, I am referring to doing round-trip tests and/or test that simply check that you actually can't hit those `unwrap`s.

D4nte:
> All this floating point arithmetic is really scary :/

I don't understand what you are talking about. This PR moves most arithmetic to unsigned integer.

luckysori:
> I considered it but the value we receive from Kraken may not be a rational fraction or ratio.

`num_rational` does provide https://docs.rs/num-rational/0.3.0/num_rational/struct.Ratio.html#method.from_float and https://docs.rs/num-rational/0.3.0/num_rational/struct.Ratio.html?search=#method.approximate_float to deal with that. And so does the other crate.

> Also I felt it would be faster to roll out what I did then learn the limitations and usage of the quoted libs.

If you're confident writing that stuff then that is fair, but I'm certainly not very confident in my ability to review it :face_with_head_bandage:

src/publish.rs
```diff
@@ -1,20 +1,46 @@
...
+    pub buy_amount: Amount,
+}
+
+#[derive(Debug, Copy, Clone, Ord, PartialOrd, PartialEq, Eq)]
+struct Amount(u64);
+
+impl Amount {
+    #[allow(clippy::cast_precision_loss)] // It's ok because it just means we are applying slightly more than the given spread
+    #[allow(clippy::cast_possible_truncation)] // We probably want to use custom amounts down the line
+    #[allow(clippy::cast_sign_loss)] // It's ok because all values should be positive
```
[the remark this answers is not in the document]
D4nte: This is now gone.

src/publish.rs
D4nte: According to https://github.com/coblox/nectar/pull/14/files we get a float point rate from Kraken.

I applied the spread on the rate to limit the float point arithmetic and isolate it to that this unique operation: apply spread on rate.

If I were to apply the rate and then later the spread I'd end up with more float point arithmetic.

src/dai.rs
D4nte: `A type for lazy initialization of e.g. global static variables, which provides the same functionality as the lazy_static! macro.`

Looks the same to me.

D4nte:
Will use existing crate.

src/publish.rs
thomaseizinger: > According to [#14 (files)](https://github.com/coblox/nectar/pull/14/files) we get a float point rate from Kraken.

Not quite, Daniel calculates it from responses like these: https://github.com/coblox/nectar/pull/14/files#diff-c4a4631dcbd8719f32d727ef48b9ce7fR117-R144

src/dai.rs
thomaseizinger: I personally prefer it because it doesn't use macros but only regular Rust syntax, meaning I don't have to remember exactly what syntax the macro-author decided on :)

src/publish.rs
thomaseizinger: > If I were to apply the rate to get the buy amount and then apply the spread I'd end up with 2 operations doing float point arithmetic.

Not if the input is already the buy-amount and you just need to apply the spread.

That is also just a multiplication then we doesn't need to involve floating point arithmetic.

D4nte: > > According to [#14 (files)](https://github.com/coblox/nectar/pull/14/files) we get a float point rate from Kraken.
> 
> Not quite, Daniel calculates it from responses like these: [#14 (files)](https://github.com/coblox/nectar/pull/14/files#diff-c4a4631dcbd8719f32d727ef48b9ce7fR117-R144)

The link does not seem to work. I see floating point values in there:
```
    fn ohlc_with_vwap() -> Ohlc {
        // These are actual values from Kraken
        Ohlc {
            timestamp: Utc::now(),
            high: 9825.1,
            low: 9791.0,
            vwap: 9806.7,
            trading_pair: TradingPair::BtcDai,
        }
    }
```

It's a rate but it is a price.

D4nte:
`BigRational` is [not an option](https://github.com/coblox/nectar/pull/17) because it loses precision.

D4nte:
> Can we discuss the idea of not converting to a floating point rate at all?

I do not understand how you intend to manipulate decimal numbers without using floating point representations.

We have two requirements that uses decimal numbers:
- The value/rate from Kraken.
- Applying a spread.

Please write the (pseudo) code to demonstrate how you would implement this requirements without dealing with floating points.

src/publish.rs
D4nte: > That is also just a multiplication then we doesn't need to involve floating point arithmetic.

Are you saying that multiplying floating points does not involve floating point arithmetic?

> ask for a the rate from kraken, we get 1BTC = 9543 DAI
> apply a spread to the amount

How do apply a spread to this amount without doing float point arithmetic?

> That is also just a multiplication then we doesn't need to involve floating point arithmetic.

It's a multiplication that involves decimal, so yeah, it does involves floating point arithmetic.

I think you should make your point across via code.

thomaseizinger:
> > Can we discuss the idea of not converting to a floating point rate at all?
> 
> I do not understand how you intend to manipulate decimal numbers without using floating point representations.
> 
> We have two requirements that uses decimal numbers:
> 
>     * The value/rate from Kraken.
> 
>     * Applying a spread.
> 
> 
> Please write the (pseudo) code to demonstrate how you would implement this requirements without dealing with floating points.

To my understanding, the "don't use floating point" advice applies to doing multiplication and other arithmetic. We can't do anything about the fact that kraken sends us a floating point value.

What we can do though is limit the usage as soon as possible, use datatypes that store exact values (u64 or u128) and use sub-cent accounting to minimize the impact of precision loss in a controlled way.

Kraken gives the amount of DAI for 1 BTC. DAI is basically USD, so it comes with 2 decimal places of precision.

First, we have to decide on how accurate we want to be in terms of cents, let's assume that "millicents" are good enough.

1. Multiply the value we get from kraken by 10 * 1000 (10 to get cents, 1000 to get millicents). Store this value in a u64.
2. To apply a 3% spread on that value, we:
	- divide the u64 value by 100 to get 1%.
	- multiply the result by 3
	- add the value to the original amount

If we want sub-percent spreads, we need to use per mille instead of per cent, potentially go down to microcents etc.

We do lose precision here as well, simply because there will always be divisions what have a a remainder. To my understanding, the important part is that we use datatypes that store exact values and do sub-cent accounting to make sure the precision-loss is controlled and predictable.

D4nte:
I initially asked for the review to be focused on the `WorthIn` part but instead the comments were made on code that was not yet ready (the PR was initially marked as draft) and still using float arithmetic.

The approach explained above is exactly what I was doing from day one.

I shall continue on the initial approach and will mark the PR ready once everything is done.

D4nte:
Will open fresh one once done.

PR #17, opened 2020-06-17
Use BigRational

D4nte:
Using `num::BigRational` is actually not an option because there is hidden precision loss using the crate:

```
thread 'publish::tests::given_a_rate_and_spread_return_order_with_both_amounts' panicked at 'assertion failed: `(left == right)`
  left: `103000000000000005607`,
 right: `103000000000000000000`', src/publish.rs:252:9
```

As we are managing money, this is not acceptable. Back to rolling it out myself with https://github.com/coblox/nectar/pull/16


src/publish.rs
```diff
@@ -1,185 +1,254 @@
...
-        // ? Buy => 1000 sell
-        // 100 Buy => 1000 Sell
-        // 3% spread
-        // 103 Buy => 1000 Sell
+        let order = new_dai_bitcoin_order(wallet, book, btc(9999.0), rate, spread);
 
-        assert_eq!(order.sell_amount, 1000);
-        assert_eq!(order.buy_amount, 104); // Rounding up taking in account precision loss
+        assert_eq!(order.sell_amount, btc(1000.0));
+        assert_eq!(order.buy_amount, dai(103.0));
```
luckysori, line 252: I'm probably not getting it, but from your test output and this assertion, the value you provide is less precise than the value returned by your function. How did you calculate the expected value?

thomaseizinger:
Removed myself because I think this is stale.

luckysori:
Should we close this one, @D4nte?

PR #19, opened 2020-06-18
Allow and prioritise refund during swap execution

luckysori:
I'm happy to move forward with this simpler design. Let me know if the logic is sound.

I'm not sure how this new design would fit in with two watch-only actors (cnd's use-case), but I think it's okay for Nectar.

### DIFF
--- a/src/swap.rs
+++ b/src/swap.rs
@@ -3,9 +3,10 @@
 use chrono::NaiveDateTime;
 use comit::{
     btsieve::{ethereum::ReceiptByHash, BlockByHash, LatestBlock},
-    ethereum, Secret, SecretHash,
+    ethereum, Secret, SecretHash, Timestamp,
 };
 use futures::future::{self, Either};
+use std::sync::Arc;
 
 pub mod hbit {
     use bitcoin::{secp256k1::SecretKey, *};
@@ -98,6 +99,7 @@ pub mod hbit {
         }
     }
 }
+
 pub mod herc20 {
     use chrono::NaiveDateTime;
     pub use comit::{
@@ -173,10 +175,10 @@ pub mod herc20 {
     }
 }
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Debug)]
 struct WatchOnlyAlice<AC, BC> {
-    alpha_connector: AC,
-    beta_connector: BC,
+    alpha_connector: Arc<AC>,
+    beta_connector: Arc<BC>,
     secret_hash: SecretHash,
     start_of_swap: NaiveDateTime,
 }
@@ -192,7 +194,8 @@ where
 {
     async fn fund(&self, params: &hbit::Params) -> anyhow::Result<hbit::CorrectlyFunded> {
         let event =
-            hbit::watch_for_funded(&self.alpha_connector, &params, self.start_of_swap).await?;
+            hbit::watch_for_funded(self.alpha_connector.as_ref(), &params, self.start_of_swap)
+                .await?;
 
         Ok(event)
     }
@@ -212,9 +215,12 @@ where
         _params: &herc20::Params,
         deploy_event: herc20::Deployed,
     ) -> anyhow::Result<herc20::Redeemed> {
-        let event =
-            herc20::watch_for_redeemed(&self.beta_connector, self.start_of_swap, deploy_event)
-                .await?;
+        let event = herc20::watch_for_redeemed(
+            self.beta_connector.as_ref(),
+            self.start_of_swap,
+            deploy_event,
+        )
+        .await?;
 
         Ok(event)
     }
@@ -239,7 +245,7 @@ where
         SC: bitcoin::secp256k1::Signing,
     {
         let event = hbit::watch_for_refunded(
-            &self.alpha_connector,
+            self.alpha_connector.as_ref(),
             &params,
             fund_event.location,
             self.start_of_swap,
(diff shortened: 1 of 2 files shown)


src/swap.rs
```diff
@@ -281,20 +404,54 @@ async fn hbit_herc20<A, B, SC>(
...
     SC: bitcoin::secp256k1::Signing,
 {
+    if !alice.is_safe_to_fund(herc20_params.expiry).await? {
+        return Ok(());
+    }
+
     let hbit_funded = alice.fund(&hbit_params).await?;
 
+    if !bob.is_safe_to_fund(herc20_params.expiry).await? {
+        alice.refund(&hbit_params, hbit_funded, secp).await?;
```
thomaseizinger, line 418: I am not completely sure why but it is a bit weird that we are asking Bob here before Alice does her refund.

I guess it is fine because it is the same process so the other actor is not gonna be malicious :)

src/swap.rs
```diff
@@ -321,20 +478,55 @@ impl herc20::RedeemAsAlice
...
 }
 
-#[derive(Clone, Copy, Debug)]
+#[async_trait::async_trait]
+impl hbit::Refund for WalletAlice<BitcoinWallet, EthereumWallet, hbit::PrivateDetailsFunder> {
+    async fn refund<SC>(
+        &self,
+        params: &hbit::Params,
+        fund_event: hbit::CorrectlyFunded,
+        secp: &bitcoin::secp256k1::Secp256k1<SC>,
```
D4nte, line 497: Not sure what feels the cleanest but this is the binary application, I would not mind if secp is directly used from crate root instead of passing it around as an argument.

If we really want to pass it around as an argument then we need to remove the `pub` (and even `static` keyword from its declaration in `main.rs` and `lib.rs` and initialise it from the test/bin start.

src/swap.rs
```diff
@@ -273,6 +308,94 @@ impl hbit::RedeemAsBob for WalletBob<BitcoinWallet, EthereumWallet, hbit::Privat
...
+
+#[async_trait::async_trait]
+impl<AC, BC> SafeToFund for WatchOnlyAlice<AC, BC>
+where
+    BC: LatestBlock<Block = ethereum::Block>,
+    AC: Send + Sync,
+{
+    async fn is_safe_to_fund(&self, beta_expiry: Timestamp) -> anyhow::Result<bool> {
+        let ethereum_time = ethereum_latest_time(self.beta_connector.as_ref()).await?;
+        // TODO: Apply a buffer depending on the blocktime and how
```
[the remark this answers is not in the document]
thomaseizinger: > comit will tell you whether it's time to refund or time to redeem as it should now return only one action at a time.

Cnd will tell you but `comit` doesn't know about actions (yet).
The tricky part is finding a model that works several clients here.

All the abstractions that we currently built in `comit` are tailored around that the consuming Rust client will reason about what action should be done but not actually do it.
What we are trying to do here is not that different actually.

Interestingly though, the states and events abstractions we have in `cnd` suddenly feel pretty pointless if you could in theory just call through to the wallet and just do the action right where you know you want to do it :)

> If we want to push the safety further by adding some buffer than I think it should be recommended by the comit library directly. @thomaseizinger thoughts?

Agreed but it seems to early to move all this code in the `comit` library. My recommendation would be to keep it in `nectar` for now. Use what is useful from the `comit` library, iterate and see where we end up.

Once `nectar` is done, it should be easier to see what can be consolidated into `comit` from `cnd` and `nectar`.

src/swap.rs
thomaseizinger: You are right.

I think @luckysori intention was to make this code already portable into the `comit` lib. Not sure we need to bother with that though at this stage :)

src/swap.rs
luckysori: > Alice should refund as soon as she can refund, when it's past Alpha expiry.

This refund method waits until refund is possible and then refunds. Maybe the name could reflect that somehow.

>  Bob should fund before beta expiry, if beta expiry is passed Bob should refund or not fund.

I think that's what we're doing here. A source of confusion may be that we're using `is_safe_to_fund` for both deploy and fund, for simplicity. I can split them up.

> As we work with absolutely expiries there is no safe to fund notion for Bob.
safe to fund for bob sounds like Bob would try to fund after beta expiry but not too close to alpha expiry.

You are right, the name of this function does not work here. 

Here the logic is: if Beta has expired, end the protocol. For Alice because she cannot safely reveal the secret if Bob could refund just before and learn the secret for free; for Bob because the other party should not continue, so we may as well abort and not waste money on fees for a swap that should never succeed.

Maybe we could just use `{alpha, beta}_has_expired` instead of the not always correct `is_safe_to_fund`.

src/swap.rs
```diff
@@ -281,20 +404,54 @@ async fn hbit_herc20<A, B, SC>(
...
+
+        return Ok(());
+    }
+
     let _herc20_funded = bob.fund(&herc20_params, herc20_deployed.clone()).await?;
 
-    let herc20_redeemed = alice.redeem(&herc20_params, herc20_deployed).await?;
+    if !alice.is_safe_to_redeem(herc20_params.expiry).await? {
+        alice.refund(&hbit_params, hbit_funded, secp).await?;
+        bob.refund(&herc20_params, herc20_deployed.clone()).await?;
```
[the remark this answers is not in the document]
luckysori: > Or we are just still in the phase of doing a full swap as a PoC?

We're modelling both actors here, correct. In Nectar we would model Bob as an actor that can perform wallet actions (see `WalletBob`) and Alice as a watcher for Alice's actions (see `WatchOnlyAlice`).

src/swap.rs
luckysori: Yep, can revert this!

src/swap.rs
D4nte: Yes I'd be keen for a rename and just use expiries as time markers to decide.

PR #23, opened 2020-06-24
Change rate logic to use Kraken ticker instead of OHLC endpoint

da-kami:
After @bonomat pointed out that we should use the ticker's ask and bids for coming up with the rate rather than the OHLC I implemented with quickly.

Reasoning:

This is important, because the `OHLC` depicts a market price that is in the past, whereas the `ask` and `bid` represent the current _expected_ price in the market. 

I did several fast iterations on that thing. Fast code churn style.

Since the `Rate` model was not optimal yet I iterated on that and introduced `kraken::btc_dai::Rate` and `kraken::btc_dai::MidMarketRate`.
Removed `markets`, removed `TradingPair`, removed `Position`.

Note that we might want to consider re-introducing that `Position` for the 
More on that in blogpost on basecamp - coming soon :)

Note that some clippy stuff was fixed because the errors/warning were annoying.


src/markets.rs
```diff
@@ -20,124 +21,36 @@ pub enum Position {
...
+pub async fn get_rate(trading_pair: TradingPair, position: Position) -> anyhow::Result<Rate> {
 
-        let trading_pair = self.trading_pair;
-        let timestamp = self.timestamp;
+    let ask_bid = kraken::get_ask_bid(trading_pair).await?;
 
-        match position {
-            Position::Buy => Ok(Rate {
+    match position {
+        // ask = sell
```
da-kami, line 30: this is confusing... I feel we should come up with a better model for this :)

da-kami:
@D4nte @luckysori note that I also fixed a bunch of unrelated clippy errors/warnings (in tests and elsewhere) with this now, because it was quite annoying when trying to figure out what is still to be fixed.

src/bitcoin.rs
```diff
@@ -33,7 +33,7 @@ impl Amount {
 
     /// Allow to know the worth of self in dai using the given conversion rate.
     /// Truncation may be done during the conversion to allow a result in attodai.
-    pub fn worth_in(&self, btc_to_dai_rate: Rate) -> dai::Amount {
+    pub fn worth_in(self, btc_to_dai_rate: Rate) -> dai::Amount {
```
D4nte, line 36: Why consuming if you can avoid it? Or you cannot avoid it anymore?

src/kraken.rs
```diff
@@ -0,0 +1,139 @@
...
+    bid: Vec<String>,
+}
+
+impl TryFrom<TickerData> for Rate {
+    type Error = serde_json::Error;
+
+    fn try_from(value: TickerData) -> Result<Self, Self::Error> {
+        let ask_price = value
+            .ask
+            .first()
```
D4nte, line 43: Is `first` (returned by Kraken) the most recent or the highest? We are looking for the highest price, aren't we?

src/kraken.rs
```diff
@@ -0,0 +1,139 @@
...
+            .ok_or_else(|| serde_json::Error::custom("no ask price"))?;
+        let bid_price = value
+            .bid
+            .first()
+            .ok_or_else(|| serde_json::Error::custom("no bid price"))?;
+
+        Ok(Rate {
```
D4nte, line 50: Ok yeah, this is not a `Rate`. You could call it `BidAndAsk`: https://www.investopedia.com/terms/b/bid-and-ask.asp

src/kraken.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+pub mod btc_dai {
+    use super::*;
+    use chrono::{DateTime, Utc};
+
+    #[derive(Debug, Copy, Clone, PartialEq)]
+    pub struct MidMarketRate {
+        value: f64,
+        timestamp: DateTime<Utc>,
```
D4nte, line 68: Do we ever use the timestamp?

src/rate.rs
```diff
@@ -83,7 +83,7 @@ impl Spread {
         Ok(Spread(permyriad))
     }
 
-    pub fn apply(&self, rate: Rate) -> anyhow::Result<Rate> {
+    pub fn apply(self, rate: Rate) -> anyhow::Result<Rate> {
```
D4nte, line 86: Why this change?

luckysori:
@D4nte: I'm going to merge this as is and iterate on it keeping your comments in mind.

src/kraken.rs
luckysori: AFAIK, the ask field of a trading ticker contains the latest information. Same for the bid.

Here `.first()` is just selecting the first field in the array `[price, whole_lot_volume, lot_volume]`, according to https://www.kraken.com/features/api.

src/kraken.rs
luckysori: @da-kami: what was the plan for the timestamp? I am keen to remove it.

PR #25, opened 2020-06-25
Implement wallet actions and verify that swap::hbit_herc20 works

luckysori:
This PR finally allows Nectar to execute Hbit<->Herc20 swaps in the role of Bob. This is tested by an integration test, which drove forward the development of the Bitcoin and Ethereum wallets. Keep in mind that the wallets have been developed in a minimal way to fulfil the requirements of the test, so we'll very likely need to do more work after this.


luckysori:
> I assume that we eventually will add refund tests as well to see that the refunding logic works as well (is glued together correctly).

Yes, we should test the other paths. I wonder what would be the best strategy for that. It seems trivial to make it exit straight away, but the intermediate paths appear harder to hit.

src/ethereum_wallet.rs
```diff
@@ -0,0 +1,209 @@
...
+
+        let geth_client = geth::Client::new(url);
+
+        Ok(Self {
+            private_key,
+            geth_client,
+        })
+    }
+
+    pub fn new_from_private_key(private_key: clarity::PrivateKey, url: Url) -> Self {
```
D4nte, line 28: if that's only used in test it'd be good to have it behind `#[cfg(test)]` if possible

src/ethereum_wallet.rs
```diff
@@ -0,0 +1,209 @@
...
+            private_key,
+            geth_client,
+        }
+    }
+
+    pub fn account(&self) -> anyhow::Result<Address> {
+        let address = self
+            .private_key
+            .to_public_key()
+            .map_err(|_| anyhow::anyhow!("Failed to derive public key from private key"))?;
```
D4nte, line 41: Can it really fail? I'd be keen to remove the `Result` return.

src/ethereum_wallet.rs
```diff
@@ -0,0 +1,209 @@
...
+            gas_limit,
+            chain_id,
+            ..
+        }: DeployContract,
+    ) -> anyhow::Result<Hash> {
+        let nonce = self.get_transaction_count().await?;
+
+        let transaction = clarity::Transaction {
+            nonce: nonce.into(),
+            gas_price: 0u32.into(),
```
D4nte, line 59: What happen when gas price is 0?

src/ethereum_wallet.rs
```diff
@@ -0,0 +1,209 @@
...
+            gas_limit: gas_limit.into(),
+            to: clarity::Address::default(),
+            value: 0u64.into(),
+            data: data.0,
+            signature: None,
+        };
+
+        #[allow(clippy::cast_possible_truncation)]
```
D4nte, line 67: where is the truncation? Best if it can be avoided (this function returns a `Result` so use it) or at least explained.

src/ethereum_wallet.rs
```diff
@@ -0,0 +1,209 @@
...
+                ))?)
+            );
+
+        let hash = self
+            .geth_client
+            .send_raw_transaction(transaction_hex)
+            .await?;
+
+        // TODO: Return TransactionReceipt instead
+        std::thread::sleep(std::time::Duration::from_millis(2000));
```
D4nte, line 84: Quick todo? I see you already have a function to retrieve receipt.

src/geth.rs
```diff
@@ -0,0 +1,197 @@
...
+    ethereum::{Address, Bytes, ChainId, Hash, Transaction, TransactionReceipt},
+};
+use ethereum_types::U256;
+use num::{BigUint, Num};
+use serde_hex::{CompactPfx, SerHex};
+
+#[derive(Debug, Clone)]
+pub struct Client {
+    rpc_client: jsonrpc::Client,
+    jsonrpc_version: String,
```
D4nte, line 15: This is a constant can it be declared at such? (scope can be this module or the impl block of Client).

src/geth.rs
```diff
@@ -0,0 +1,197 @@
...
+        Client {
+            rpc_client: jsonrpc::Client::new(url),
+            jsonrpc_version: String::from("2.0"),
+        }
+    }
+
+    pub async fn chain_id(&self) -> anyhow::Result<ChainId> {
+        let chain_id = self
+            .rpc_client
+            .send::<Vec<()>, String>(jsonrpc::Request::new(
```
D4nte, line 29: `String`: geth is the worst.

src/geth.rs
```diff
@@ -0,0 +1,197 @@
...
+                self.jsonrpc_version.clone(),
+            ))
+            .await
+            .context("failed to get transaction receipt")?;
+
+        Ok(receipt)
+    }
+
+    pub async fn get_transaction_count(&self, account: Address) -> anyhow::Result<u32> {
+        let count: String = self
```
D4nte, line 105: [This should return a number,](https://web3js.readthedocs.io/en/v1.2.7/web3-eth.html#gettransactioncount) it does not?

src/main.rs
```diff
@@ -1,39 +1,3 @@
-#![warn(
```
D4nte, line 1: Are you sure about that?

src/seed.rs
```diff
@@ -0,0 +1,34 @@
+use bitcoin::{secp256k1, secp256k1::constants::SECRET_KEY_SIZE};
```
D4nte, line 1: Sorry about the merge conflict!

src/swap.rs
```diff
@@ -415,397 +31,112 @@ where
     let hbit_funded = alice.fund(&hbit_params).await?;
 
     if !bob.is_safe_to_fund(herc20_params.expiry).await? {
```
D4nte, line 33: I thought you were ok to rename those functions?

src/swap.rs
```diff
@@ -415,397 +31,112 @@ where
...
-}
-
-async fn ethereum_latest_time<C>(connector: &C) -> anyhow::Result<Timestamp>
-where
-    C: LatestBlock<Block = ethereum::Block>,
-{
-    let timestamp = connector.latest_block().await?.timestamp.into();
-
-    Ok(timestamp)
+pub trait SafeToRedeem {
```
D4nte, line 89:
```suggestion
pub trait IsSafeToRedeem {
```

D4nte: As previously discussed, I don't think `safe` is the correct terminology.
To me, it's `it's not time to refund` would be a better descriptor of the behaviour I expect.

src/seed.rs
```diff
@@ -0,0 +1,34 @@
...
+    }
+}
+
+impl Default for Seed {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl From<[u8; SECRET_KEY_SIZE]> for Seed {
```
D4nte, line 30: If that's only for test, please put it behind `#[cfg(test)]`

src/swap/alice.rs
```diff
@@ -0,0 +1,262 @@
...
+
+#[async_trait::async_trait]
+impl<AC, BC> SafeToFund for WatchOnlyAlice<AC, BC>
+where
+    BC: LatestBlock<Block = ethereum::Block>,
+    AC: Send + Sync,
+{
+    async fn is_safe_to_fund(&self, beta_expiry: Timestamp) -> anyhow::Result<bool> {
+        let ethereum_time = ethereum_latest_time(self.beta_connector.as_ref()).await?;
+        // TODO: Apply a buffer depending on the blocktime and how
```
D4nte, line 98: I'd rather remove the todo and rename the function at this stage. The "buffer" should be a properly discussed subject out of scope of the nectar project because it's not that straightforward.
I actually had a shaped idea around that, need to discuss with @bonomat.

src/bitcoin_wallet.rs
```diff
@@ -1,51 +1,25 @@
...
     name: String,
     bitcoind_client: bitcoind::Client,
-    private_key: ::bitcoin::PrivateKey,
+    private_key: bitcoin::PrivateKey,
 }
 
 impl Wallet {
-    pub fn new(seed: Seed, url: Url, network: Network) -> anyhow::Result<Wallet> {
-        let private_key = ::bitcoin::PrivateKey {
+    pub fn new(seed: Seed, url: Url, network: Network, name: String) -> anyhow::Result<Wallet> {
```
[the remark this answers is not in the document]
luckysori: It was just convenient for debugging tests. I can revert this. Should I reintroduce your random name derived from the seed?

src/bitcoin_wallet.rs
```diff
@@ -90,18 +64,52 @@ impl Wallet {
...
-    fn gen_name(private_key: PrivateKey) -> String {
-        let mut hash_engine = PubkeyHash::engine();
-        private_key
-            .public_key(&crate::SECP)
-            .write_into(&mut hash_engine);
+    pub async fn send_to_address(
+        &self,
+        address: Address,
+        amount: Amount,
+        network: Network,
```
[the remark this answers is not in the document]
luckysori: I guess for Bitcoin that would work well, because it's already part of the node which belongs to a single network.

src/ethereum_wallet.rs
luckysori: It works on a local development node, but it would never get mined on mainnet. Should I use different values depending on the network? Should it be a constant?

src/geth.rs
luckysori: It's 1.0 for bitcoind and 2.0 for geth :(

src/geth.rs
luckysori: It returns a hex string :)))

src/swap.rs
luckysori: I have gone for a slightly different approach on a separate branch which builds on top of this. These traits will be gone! ☺️

src/swap/alice.rs
luckysori: I tend to agree with you. I'll remove the TODOs.

src/bitcoin_wallet.rs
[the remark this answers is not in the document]
D4nte: Either way, for Ethereum, it should also be a field of the wallet, I am not saying that we should not check the network at each call but that once a wallet is instantiated, the network should not change.

src/ethereum_wallet.rs
D4nte: Well, I think it's more a proper todo (task in basecamp) to decide on network fees using an external service.

src/geth.rs
D4nte: Yes, then declare a constnant in `geth.rs` and another one in `bitcoind.rs`

src/geth.rs
D4nte: Maybe a todo here to actually cast it to a number using some hex deserialisation?

src/bitcoin_wallet.rs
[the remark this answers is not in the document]
luckysori: > Should I reintroduce your random name derived from the seed?

I did.

src/ethereum_wallet.rs
luckysori: Nice catch! Must have been from an old version.

src/geth.rs
luckysori: The casting happens a few lines down:

```rust
let count = u32::from_str_radix(&count[2..], 16)?;
```

But eventually we could/should remove the deserialization logic from this function.

src/main.rs
luckysori: Pretty sure!

luckysori:
@D4nte: I'm going ahead with the merge. You can keep commenting here and I'll keep it all in mind for the next PR.

PR #26, opened 2020-06-25
PoC for main loop

da-kami:
Thanks to @thomaseizinger for helping me fleshing this out :)

The idea cycles around one main event loop, that first always updates the mid-market-rate and balances in the `Maker` and then reacts to events happening on the network. Reacting to network events has a timeout, so we ensure that we update the balance and rate periodically.
(Additionally the maker's logic can actually ensure that rate/balance is not outdated from his point of view if e.g. we cannot fetch rate from Kraken no orders should be published until we have an up-to-date rate again - but that is future music :)

Events that can happen on the network: 
- Order expired 
- Order taken 
- Swap finalized

The `Maker`, the central component that bundles the decision logic, holds all the states and reacts to events happening on the network.
After handling an event the `Maker` returns an `Action` (that includes possible data, e.g. a new `Order`) that can be actioned on by the event loop (i.e. publishing the new order in the orderbook). The event -> action model helps us with testing the Maker's logic - see the POC test in `main`.

Create this as a Draft for it now - we can also merge it if we feel like it.


src/main.rs
```diff
@@ -34,6 +46,327 @@ pub mod test_harness;
...
+
+                unimplemented!()
+            }
+            NetworkEvent::SwapFinalized(local_swap_id, remote_date) => {
+                // TODO: Bookkeeping, lock up funds
+                Some(Action::StartSwapExecution(local_swap_id, remote_date))
+            }
+            _ => None,
+        }
+    }
```
thomaseizinger, line 206: You might want to decouple the `Maker` from the `NetworkEvent` by simply providing functions that are equivalent to the individual variants of the enum, i.e in the loop, do:

```rust
match network_event {
	NetworkEvent::OrderExpired(order) => maker.handle_order_expired(order)
}
```
That way, the maker doesn't need to care about how he learns that the order is expired (i.e. from the network).

We already do the same thing with the action. We match on it an call functions on the appropriate other components and not pass the action to the component :)

thomaseizinger: Thinking more about this led me to the following - I think more consistent - design:

1. Give the maker "command" like functions similar to what is outlined above:

`maker.expire_order()`

2. Make each of those functions return an `Event` about what happened as a result (and not an action):

`fn expire(&mut self) -> MakerEvent`

```
enum MakerEvent {
    OrderCreated
}
```

The reasoning is that if the Maker is supposed to be in charge of everything, then the functions should be commands. Those can then potentially fail by returning a `Result` for example. Sounds useful for `Maker.take_order`.
Generally, it decouples unrelated functionality from each other, which should simplify the type signature and the implementation.

Regarding Event vs Action:

After we have executed a command, the state already changed inside the maker.
So the fact that a new order was created and the maker did the accounting for it is not an action but describes something in the past. In other words: an event.

The fact that we should publish an order after it has been created is actually not that important.
Within the test, it doesn't make sense for example because the test can't actually follow the action :)

da-kami: > Those can then potentially fail by returning a Result for example. Sounds useful for Maker.take_order.

That is something we will need - otherwise a "generic" action is needed that is either "actually take and execute" or "don't take and report back to counterparty. 

I will give your suggestion a try and see how it feels. Even though I like the idea of `Event -> Action` (an event on the network leads to an action) that does not really go together well and I agree that the Maker would emit events. (We actually had this in one of the iterations, but not exactly like outlined :)

src/main.rs
```diff
@@ -34,6 +46,327 @@ pub mod test_harness;
...
+    SwapFinalized(LocalSwapId, RemoteData),
+}
+
+#[derive(Debug, PartialEq)]
+pub enum Action {
+    PublishNewOrder(Order),
+    StartSwapExecution(LocalSwapId, RemoteData),
+}
+
+// Bundles the state of the application
```
D4nte, line 142: The state at which point? I am not sure this is the best representation for the balances, let's see how it all fit together.

src/main.rs
```diff
@@ -34,6 +46,327 @@ pub mod test_harness;
...
+    btc_fee: bitcoin::Amount,
+    dai_fee: dai::Amount,
+    btc_locked_funds: bitcoin::Amount,
+    dai_locked_funds: dai::Amount,
+    btc_max_sell_amount: bitcoin::Amount,
+    dai_max_sell_amount: dai::Amount,
+    rate: MidMarketRate,
+    spread: Spread,
+
+    active_orders: u32,
```
D4nte, line 155: Shouldn't this be a container of actual active orders?

src/main.rs
```diff
@@ -34,6 +46,327 @@ pub mod test_harness;
...
+            }
+            _ => None,
+        }
+    }
+
+    //
+    pub fn update_rate(&mut self, new_rate: ()) {}
+
+    // the balance is to be updated once the trade was actually setteled, i.e. the swap execution is finished
+    pub fn update_bitcoin_balance(&mut self, balance: bitcoin::Amount) {
```
D4nte, line 212: I am not sure exactly what is the intent here. I would expect to indeed have a function that allows you and update and get the `bitcoin_locked_funds` for trades taken but not yet funded (and actually, this value should be updated once a trade is funded).
Regarding the actual Bitcoin balance, we need to ask the wallet every time. Maybe the user withdrew or deposited funds.

src/main.rs
```diff
@@ -34,6 +46,327 @@ pub mod test_harness;
...
+        self.btc_balance = balance;
+    }
+
+    pub fn track_failed_rate(&mut self, error: anyhow::Error) {}
```
D4nte, line 216: Do we really need this at this stage?

src/main.rs
```diff
@@ -34,6 +46,327 @@ pub mod test_harness;
...
+                        }
+                    }
+                }
+            }
+            _ => (),
+        }
+    }
+}
+
+// TODO: Shouldn't the wallet return our own bitcoin amount?
```
D4nte, line 269: Yes it should :)

src/main.rs
da-kami: Most likely yes :) 
It is represented as something very basic at the moment one can take a decision on - it actually remained in there from a previous iteration that implemented a test for an "initial order". 

I will remove it and introduce a proper `Order` store once we actually start building that logic.

src/main.rs
```diff
@@ -34,6 +46,327 @@ pub mod test_harness;
 pub static SECP: Lazy<::bitcoin::secp256k1::Secp256k1<::bitcoin::secp256k1::All>> =
     Lazy::new(::bitcoin::secp256k1::Secp256k1::new);
 
-fn main() {
-    println!("Hello, world!");
+/// A `NetworkBehaviour` that delegates to the `Announce`, `Orderbook`, `Take` and `Comit` (execution) behaviours.
+#[derive(NetworkBehaviour)]
+#[behaviour(out_event = "NetworkEvent", poll_method = "poll")]
+#[allow(missing_debug_implementations)]
+pub struct Nectar {
```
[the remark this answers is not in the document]
da-kami: I moved it, also moved the orderbook (currently a placeholder) to a separate module.
I kept the `Maker` in `main.rs` for the moment, but moved it into a `maker` module there.

src/main.rs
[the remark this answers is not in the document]
da-kami: I think it will depend on if the `Maker` wants to use this information as part of his decision-making. If so, I'd say it should remain in the maker as part of its state. 

@D4nte we can think this through tomorrow and take a decision.

src/main.rs
```diff
@@ -34,6 +44,219 @@ pub mod test_harness;
...
-fn main() {
-    println!("Hello, world!");
+mod maker {
+    use super::*;
+    use comit::LocalSwapId;
+
+    #[derive(Debug, PartialEq)]
+    pub enum New_Order {
+        Created(Order),
+    }
```
thomaseizinger, line 54: Why isn't this just a newtype? :)

```rust
pub struct OrderCreated(Order);
```

If you need an enum with more events at some points, you can always do this:

```rust
enum Event {
	OrderCreated(OrderCreated)
}
```

src/main.rs
```diff
@@ -34,6 +44,219 @@ pub mod test_harness;
...
+        pub fn expire_order(&mut self, order: Order) -> anyhow::Result<maker::New_Order> {
+            // TODO: Bookkeeping, decisionmaking if new order
+
+            // TODO: creating a new order should not take the wallet / book
+            // let new_order = new_dai_bitcoin_order();
+            let new_order = Order {
+                expiration_timestamp: Utc::now(),
+            };
+
+            // Why would we need an Event here anyway?
```
thomaseizinger, line 100: I would say to be more descriptive of what the returned order actually is.

da-kami:
closing and re-opening as actual PR

PR #28, opened 2020-06-26
Iteration on mid-market rate and order design

luckysori:
Do you guys think this helps at all?


src/order.rs
```diff
@@ -0,0 +1,247 @@
...
+
+#[derive(Debug, Clone, Copy)]
+pub struct Buy;
+
+#[derive(Debug, Clone, Copy)]
+pub struct Sell;
+
+impl BtcDaiOrder<Buy> {
+    pub fn new<W, B>(
+        wallet: W,
```
da-kami, line 36: I think this can actually be the bitcoin wallet in this case, because we are producing a specific `BtcDai<Buy>` order here.

luckysori: Totally agree.

src/order.rs
```diff
@@ -0,0 +1,247 @@
+//! The maker creates an order that defines how much he wants to buy for the amount he is selling.
```
da-kami, line 1: Would remove this comment - I wrote most of it at some point but (especially with changing the model) it is more confusing than in helps.

src/order.rs
da-kami, line 36: It might actually be better to just pass the amounts here (instead of the full wallet).
The way main event look is currently designed the wallet might not be passed around.

The wallet is a specific ethereum wallet at this point. I think we can remove this abstraction for now and make it simpler by just passing in the Amounts directly (same for the book).

PR #31, opened 2020-07-02
Rework swap execution to enable respawn

luckysori:
_Note: I think all the patches compile and make sense, but I would just review the final state given that I changed the design throughout._

`WalletACTOR` action execution flow:
1. Have I done it before?
  - Yes -> Return corresponding event.
  - No -> Go to (2).
2. Should I do it?*
  - Yes -> Do it and return corresponding event.
  - No -> Exit the protocol.

`WatchOnlyACTOR` action execution flow:
1. Have they done it before?
  - Yes -> Return corresponding event.
  - No -> Go to (2).
2. Should they do it?*
  - Yes -> Watch for event on the blockchain and return when ready.
  - No -> Exit the protocol.

Depending on how far into the protocol we are, exiting may lead to instructing (or waiting for) Alice and Bob to refund.

This PR introduces and uses traits inspired by cnd which allow us to ask and tell a database (not yet implemented) about the events of a swap.

---
(*) It hasn't changed with this PR, but I would like to clarify that the logic is very conservative i.e. if the shortest of the two expiries, beta expiry, has been reached we abort the protocol. This is because one of the parties, Alice, _should_ abort as she cannot redeem beta asset safely, so its simpler to assume that she will abort.


src/ethereum_wallet.rs
```diff
@@ -39,7 +39,10 @@ impl Wallet {
     pub fn account(&self) -> Address {
         let address = self.private_key.to_public_key().expect("cannot fail");
```
D4nte, line 40:
```suggestion
        let public_key = self.private_key.to_public_key().expect("cannot fail");
```

src/swap/db.rs
```diff
@@ -0,0 +1,9 @@
+#[async_trait::async_trait]
+pub trait Load<T>: Send + Sync + 'static {
+    async fn load(&self, swap_id: u8) -> anyhow::Result<Option<T>>;
```
D4nte, line 3: 📣 We need a real swap id here, not a `u8`.

luckysori: Yep, yep. I wasn't sure if I should introduce it with this PR as I don't actually have one when I start the swap. I expected to modify these functions once the concept is in place (with the introduction of the database, I guess).

luckysori:
> 
> 
> Awesome PR.
> 
> Not sure I saw it or missed it, but is it tested? (the recovery part).

I decided against adding tests until we have an actual database. If we want tests at this stage I think I would have to mock a lot of things.

luckysori:
As discussed in call with @D4nte, I've added the swap ID and we can proceed with the merge. I will build on top of this PR to add tests which verify that actions are only performed once. We will have to mock the dependencies to do this.

src/swap/alice.rs
```diff
@@ -123,16 +135,103 @@ pub mod wallet_actor {
...
+
+                if bitcoin_time >= params.expiry {
+                    break;
+                }
+
+                tokio::time::delay_for(Duration::from_secs(1)).await;
+            }
+
+            let refund_event = self.refund(params, fund_event).await?;
+
```
da-kami, line 229: ❓ Is there a reason refund is not saved in the DB?

Wouldn't it be important to know when reasoning if we have to do it again on a swap?

luckysori: You're right. I have a TODO in PR #35 to deal with this :)

PR #32, opened 2020-07-03
Main event loop and maker logic

da-kami:
Since this is ever-growing, I would merge the current state and then extend it. (I am branching off this state and work on rate/balance update logic.)

Note that I added `dai_balance()` to the ethereum wallet for convenience. 

Implemented:

- Logic for handle take requests (buy and sell)

Outlined:

- Initialisation (TODOs for config plugs there)

To be added:

- Logic for update rate
- Logic for update balance


src/dai.rs
```diff
@@ -82,6 +94,17 @@ impl std::ops::Sub for Amount {
     }
 }
 
+impl From<Erc20> for Amount {
+    fn from(_erc20: Erc20) -> Self {
+        // TODO: inner BigInt is private in comit lib - what is the best way to convert?
```
da-kami, line 99: @D4nte @luckysori would you have any suggestions for this?

da-kami: Note: I fixed it by re-constucting the BigInt from bytes for the moment.

src/main.rs
```diff
@@ -1,3 +1,146 @@
...
+            swarm.orderbook.publish(sell_order.into());
+            swarm.orderbook.publish(buy_order.into());
+        }
+        _ => panic!("Unable to publish initial orders!"),
+    }
+
+    loop {
+        let rate = get_btc_dai_mid_market_rate().await;
+        match rate {
+            Ok(new_rate) => maker.update_rate(unimplemented!()), // maker should record timestamp of this
```
D4nte, line 95: `// maker should record timestamp of this` why?

src/main.rs
```diff
@@ -1,3 +1,146 @@
...
+        }
+
+        // if nothing happens on the network for 15 seconds, loop
+        // again
+
+        // ASSUMPTION: a BtcDaiOrder Sell order and a BtcDaiOrder Buy
+        // order were published before we enter the loop (during
+        // initialization)
+        #[allow(clippy::single_match)]
+        match tokio::time::timeout(Duration::from_secs(15), swarm.next()).await {
```
D4nte, line 117: Make `15` come from the config.

src/main.rs
```diff
@@ -1,3 +1,146 @@
...
+            Ok(event) => {
+                match event {
+                    network::Event::OrderTakeRequest(order) => {
+                        // decide & take & reserve
+                        let reaction = maker.react_to_taken_order(order.clone());
```
D4nte, line 122:
```suggestion
                        let result = maker.process_taken_order(order.clone());
```
I find the terminology awkward.

src/main.rs
```diff
@@ -1,3 +1,146 @@
...
+
+                        match reaction {
+                            Ok(Reaction::Confirmed { next_order }) => {
+                                swarm.orderbook.take(order);
+                                orderbook.publish(next_order.into());
+                            }
+                            Ok(Reaction::RateSucks)
```
D4nte, line 129: If we handle all these scenarios the same way then it is a hint that you do not need to differentiate them. If the maker does not execute the order then it is ok to return an error instead of disguising it in an `Ok` result.

src/maker.rs
```diff
@@ -0,0 +1,394 @@
...
+    network::Order,
+    order::{BtcDaiOrder, Position},
+    rate::Spread,
+    MidMarketRate, OngoingTakers, Rate,
+};
+use std::convert::TryFrom;
+
+#[derive(Debug, PartialEq)]
+pub enum NewOrder {
+    Created(BtcDaiOrder),
```
D4nte, line 14: Can you write a comment to hint what other variants you would expect to add later?

src/maker.rs
```diff
@@ -0,0 +1,394 @@
...
+                position: Position::Buy,
+                ..
+            } => {
+                // We are buying BTC for DAI
+                // Given an order rate of: 1:9000
+                // It is NOT profitable to buy, if the current rate is greater than the order rate.
+                // 1:8800 -> We give less DAI for getting BTC -> Good.
+                // 1:9200 -> We have to give more DAI for getting BTC -> Sucks.
+                if order_rate > current_profitable_rate {
+                    return Ok(Reaction::RateSucks);
```
D4nte, line 121:
```suggestion
                    return Ok(Reaction::RateIsNotProfitable);
```

src/maker.rs
```diff
@@ -0,0 +1,394 @@
...
+                }
+
+                let updated_dai_reserved_funds = self.dai_reserved_funds.clone() + order.quote;
```
D4nte, line 124: This is a very strange way to do the maths.
I would expect to first calculate available funds: `available_funds = balance - reserved_funds` and then comaper with the order quote `quote > available_funds`

src/maker.rs
```diff
@@ -0,0 +1,394 @@
...
+                if updated_dai_reserved_funds > self.dai_balance {
+                    return Ok(Reaction::InsufficientFunds);
```
D4nte, line 126: As per my other comment, this should be a `Err`. We are interrupting the process with returning an `Result::Ok`, it's weird.

src/maker.rs
```diff
@@ -0,0 +1,394 @@
...
+
+    pub fn update_dai_balance(&mut self, balance: dai::Amount) {
+        self.dai_balance = balance;
+    }
+
+    pub fn track_failed_rate(&mut self, error: anyhow::Error) {}
+
+    pub fn track_failed_balance_update(&mut self, error: anyhow::Error) {}
+
+    pub fn next_sell_order(&self) -> anyhow::Result<BtcDaiOrder> {
```
D4nte, line 74:
```suggestion
    pub fn sell_order(&self) -> anyhow::Result<BtcDaiOrder> {
```
```suggestion
    pub fn new_sell_order(&self) -> anyhow::Result<BtcDaiOrder> {
```

The `next` terminology is weird, it is not a strict `sequence` of orders.

src/maker.rs
```diff
@@ -0,0 +1,394 @@
...
+        self.ongoing_takers
+            .insert(order.taker)
+            .expect("already checked that we can trade");
+
+        Ok(Reaction::Confirmed { next_order })
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum Reaction {
```
D4nte, line 166: This enum is actually not needed, just use a `Result`.

src/network.rs
```diff
@@ -0,0 +1,131 @@
...
+use libp2p::swarm::{NetworkBehaviourAction, PollParameters};
+use libp2p::NetworkBehaviour;
+use std::collections::VecDeque;
+use std::task::{Context, Poll};
+
+#[derive(Debug)]
+pub enum Event {
+    // The orderbook registered that somebody wants to take a specific order.
+    // We assume that this event is emitted by the orderbook.
+    OrderTakeRequest(Order), // Emitted by orderbook
```
D4nte, line 13:
```suggestion
    TakeRequest(Order), // Emitted by orderbook
```

src/ongoing_takers.rs
```diff
@@ -0,0 +1,78 @@
+use crate::network::Taker;
+use std::collections::HashSet;
+
+// TODO: Find a better name
+#[derive(Default, Debug)]
+pub struct OngoingTakers(HashSet<Taker>);
```
D4nte, line 6:
```suggestion
pub struct PeersWithOngoingTrades(HashSet<Taker>);
```

src/ongoing_takers.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub fn insert(&mut self, taker: Taker) -> anyhow::Result<()> {
+        if self.0.contains(&taker) {
+            anyhow::bail!("taker {} is already part of ongoing takers", taker.0)
+        }
+
+        self.0.insert(taker);
+        Ok(())
+    }
+
+    pub fn cannot_trade_with_taker(&self, taker: &Taker) -> bool {
```
D4nte, line 18: [Rust's naming convention](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html) is that boolean predicates should be prefixed with `is_` or other question word. It's a good practice because the current name is confusing.
```suggestion
    pub fn has_an_ongoing_trade(&self, taker: &Taker) -> bool {
```

src/maker.rs
da-kami: @luckysori do you have strong feelings about this? I'm OK with changing it back to `new` :)

src/maker.rs
da-kami: The enum helps to make things easier to understand and readable. Yes, it is technically not needed, but I think we should also distinguish an actual `Error` from a result that we expect (even though it is not the happy path).

I would tackle this in a next iteration.

src/dai.rs
D4nte: You could use `comit::asset::Erc20` as the inner of our type.

src/maker.rs
D4nte: Are you sure we need to distinguish? :)

src/ethereum_wallet.rs
```diff
@@ -11,10 +11,11 @@ use reqwest::Url;
 pub struct Wallet {
     private_key: clarity::PrivateKey,
     geth_client: geth::Client,
+    dai_contract_adr: Address,
 }
 
 impl Wallet {
-    pub fn new(seed: Seed, url: Url) -> anyhow::Result<Self> {
+    pub fn new(seed: Seed, url: Url, dai_contract_adr: Address) -> anyhow::Result<Self> {
```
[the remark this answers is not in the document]
da-kami: I removed the parameter from the constructor, but left the field, setting it to the mainnet address and added a TODO.
Note, that we don't seem to know the `ChainId` upon initialisation, so we probably will have handle this upon calling `dai_balance()`.

Can be done in separate follow up (there seem to be other `TODO`s around this - e.g. validating that we are on the correct network - can be tackled together.)

src/main.rs
da-kami: Outdated, from previous iteration.

src/maker.rs
da-kami: It really helps when reading the tests as well... If you feel strong about it please go for follow up, I won't tackle it here. :)

src/maker.rs
da-kami: I renamed it to `new_...` for now, if we feel strong about these names we can always change them later again.

src/dai.rs
da-kami: Yeah! Will record it as follow up TODO, because it touches quite a lot of unrelated code parts.

src/ethereum_wallet.rs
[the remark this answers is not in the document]
D4nte: I did that in my PR #37

PR #33, opened 2020-07-03
Add db

D4nte:
Ready to merge

I decided to go with `sled` crate. This is the most popular embedded crate I could find, it was also considered as part of the [DB spike](https://github.com/comit-network/spikes/blob/master/0009-comit-btsieve-db.adoc).

One of the downside for sled is that it is still `alpha`. It is now considered as `beta`. As it has not yet reached `1.0` we might want to have non-regression test on the binary format to ensure that we do not upgrade sled to a non backward compatible version and corrupt our db by mistake.

Please note that as this stage I only cater for events and cannnot load a `full` swap. Let's merge as is and do the `full swap save/load` in a new PR.

@luckysori can you point where you are using the `load full swap` feature to recover a swap so I know what API I need to provide?

### DIFF
--- a/src/swap/db.rs
+++ b/src/swap/db.rs
@@ -3,6 +3,7 @@ use crate::SwapId;
 use anyhow::{anyhow, Context};
 use comit::asset::Erc20;
 use comit::ethereum::{self, Hash, Transaction, U256};
+use comit::Secret;
 use serde::{Deserialize, Serialize};
 use serde_hex::{SerHexSeq, StrictPfx};
 
@@ -73,6 +74,7 @@ struct Swap {
     pub hbit_refunded: Option<HbitRefunded>,
     pub herc20_deployed: Option<Herc20Deployed>,
     pub herc20_funded: Option<Herc20Funded>,
+    pub herc20_redeemed: Option<Herc20Redeemed>,
 }
 
 impl Default for Swap {
@@ -83,6 +85,7 @@ impl Default for Swap {
             hbit_refunded: None,
             herc20_deployed: None,
             herc20_funded: None,
+            herc20_redeemed: None,
         }
     }
 }
@@ -149,7 +152,7 @@ impl Load<hbit::Funded> for Database {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 struct HbitRedeemed {
     pub transaction: comit::transaction::Bitcoin,
-    pub secret: comit::Secret,
+    pub secret: Secret,
 }
 
 impl From<HbitRedeemed> for hbit::Redeemed {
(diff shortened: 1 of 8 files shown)


src/swap/db.rs
```diff
@@ -9,3 +16,757 @@ pub trait Load<T>: Send + Sync + 'static {
...
+            None => {
+                let mut swap = stored_swap.clone();
+                swap.hbit_funded = Some(event.into());
+
+                let old_value = serde_json::to_vec(&stored_swap)
+                    .context("Could not serialize old swap value")?;
+                let new_value =
+                    serde_json::to_vec(&swap).context("Could not serialize new swap value")?;
+
+                self.db
```
luckysori, line 161: :question: Why is `compare_and_swap` preferred over plain old `insert`?

src/swap/db.rs
```diff
@@ -9,3 +16,757 @@ pub trait Load<T>: Send + Sync + 'static {
...
+    }
+}
+
+#[async_trait::async_trait]
+impl Save<hbit::Funded> for Database {
+    async fn save(&self, event: hbit::Funded, swap_id: SwapId) -> anyhow::Result<()> {
+        let stored_swap = self.get(&swap_id)?;
+
+        match stored_swap.hbit_funded {
+            Some(_) => Err(anyhow!("Hbit Funded event is already stored")),
```
luckysori, line 151: :question: Should we really error? What if we're trying to insert exactly the same thing? I'm just not sure.

src/swap/db.rs
```diff
@@ -9,3 +16,757 @@ pub trait Load<T>: Send + Sync + 'static {
...
+impl From<herc20::Refunded> for Herc20Refunded {
+    fn from(event: herc20::Refunded) -> Self {
+        Herc20Refunded {
+            transaction: event.transaction.into(),
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl Save<herc20::Refunded> for Database {
```
luckysori, line 488: :thinking: Maybe we could hide the repetition of all the `Save` implementations behind a macro.

src/swap/db.rs
D4nte: Just using the recommended API of sled :)

src/swap/db.rs
D4nte: Let's see how it turns out but yeah, I think we should error, there is no reason to go down this path as the actions should be executed once.

src/swap/db.rs
D4nte: Let's clean up during cool down :]

PR #35, opened 2020-07-03
Idempotent* actions

luckysori:
*If the caller has implemented the memory layer (`CheckMemory` and `Remember`) correctly.

This should give us confidence that a function using the `Do` trait, such as the `hbit_herc20` one we use to execute Hbit<->Herc20 swaps, does not lead us to executing the same action twice, even if we were to call it more than once, as could be the case if we were respawning a swap.

---
All names are up for debate. Help is welcome.


src/swap/do_action.rs
```diff
@@ -0,0 +1,212 @@
...
+
+#[async_trait::async_trait]
+pub trait CheckMemory<E> {
+    async fn check_memory(&self) -> anyhow::Result<Option<E>>;
+}
+
+#[async_trait::async_trait]
+impl<E, A> CheckMemory<E> for A
+where
+    A: db::Load<E> + std::ops::Deref<Target = SwapId>,
```
thomaseizinger, line 48: This use of `Deref` is interesting because it only works based on `SwapId` being `Copy`, otherwise you couldn't go from the _reference_ you are getting from `Deref` to an owned instance.

I think `Deref` is not really supposed to be used like that :sweat_smile: 
It is mostly designed to make container types (`Mutex`, `Arc`, `Box`) work wherever a reference to the inner `T` is accepted.

An alternative design could just pass the `SwapId` as a parameter instead of referring to it implicitly through `self`?

src/swap/do_action.rs
```diff
@@ -0,0 +1,212 @@
+use crate::{swap::db, SwapId};
+use comit::Timestamp;
+
+#[async_trait::async_trait]
+pub trait Do<E>
+where
+    Self: CheckMemory<E> + ShouldAbort + Execute<E> + Remember<E>,
+    E: Clone + Send + Sync + 'static,
+    <Self as Execute<E>>::Args: Send + Sync,
```
thomaseizinger, line 9: Putting trait bounds on traits in definition sites can be a footgun because you have to restate them at every declaration.

To avoid code duplication between several impls (if that is what you are going for), you could:

1. Define the trait without bounds
2. Make a private function that implements the actual logic
3. Implement the trait through blanket impls by forwarding to the private function

src/swap/do_action.rs
```diff
@@ -0,0 +1,212 @@
...
+where
+    A: BetaLedgerTime + Sync,
+{
+    async fn should_abort(&self, beta_expiry: Timestamp) -> anyhow::Result<bool> {
+        let beta_ledger_time = self.beta_ledger_time().await?;
+
+        Ok(beta_expiry <= beta_ledger_time)
+    }
+}
+
```
thomaseizinger, line 72: This could just be a function which gets rid of the trait altogether :)

```suggestion
async fn should_abort<A>(actor: &A, beta_expiry: Timestamp) -> anyhow::Result<bool> where A: BetaLedgerTime + Sync {
    let beta_ledger_time = self.beta_ledger_time().await?;

    Ok(beta_expiry <= beta_ledger_time)
}
```

thomaseizinger: Generally, be mindful when designing things around the notion of "instances" or "objects". Often times, a function does the job equally well without adding more abstraction layers :)

thomaseizinger: In fact, you can even get rid of the 2nd trait as-well, by just having a function that takes both timestamps and compares them :)

What is the different you may ask?

It pays off to split code into "decision logic" that just depends on data and "integration logic" that wires different modules together. Decision logic that just depends on data is easy to test and "integration logic" that doesn't make any decisions (because they are delegated to decision modules) but just wires them together is hard to do wrong in Rust.

src/swap/do_action.rs
luckysori: The idea here is to only ever implement this trait once:
```rust
#[async_trait::async_trait]
impl<E, A> Do<E> for A
where
    A: CheckMemory<E> + ShouldAbort + Execute<E> + Remember<E>,
    E: Clone + Send + Sync + 'static,
    <Self as Execute<E>>::Args: Send + Sync,
{
}
```
This trait just composes those traits together in a consistent way.

src/swap/do_action.rs
luckysori: I hear you about the incorrect use of `Deref` :)

| An alternative design could just pass the SwapId as a parameter instead of referring to it implicitly through self?

I don't think this would work well because I would have to introduce the concept of a swap id in the `Do` trait, which doesn't need to know about it i.e. you could do an action for a swap without having an identifier for it.

luckysori: Would `Into` be more appropriate?

src/swap/do_action.rs
luckysori: I would be happy to split the decision logic into a function, because this happens to be the very conservative decision logic of Nectar with regards to taking actions or not. As you say, it would make sense for it to take two timestamps and to not know about where they come from, and it would be easier to test.

I would keep the trait though because it allows implementations to define their own logic.

src/swap/do_action.rs
luckysori: @thomaseizinger: after our conversation this morning I was convinced that replacing the trait with a free function was superior, but it does come with the downside that I have to add a much noisier trait bound per action to the `hbit_herc20` function (and `herc20_hbit` in the near future).

Since I don't really see harm in the trait itself and it turns out that I can use it to hide some details I am leaning towards keeping it. Maybe you have a better idea though :)

thomaseizinger: Obviously completely up to you! :)

If that is the main justification for the trait now, then I'd still question if it is such a good design if you need to hide a lot of noise behind a trait :D

Exposing yourself to this noise/pain might yield a different and potentially better design that actually reduces the noise instead of just hiding it :)

src/swap.rs
```diff
@@ -21,33 +24,43 @@ pub async fn hbit_herc20<A, B>(
     herc20_params: herc20::Params,
 ) -> anyhow::Result<()>
 where
-    A: hbit::Fund + herc20::RedeemAsAlice + hbit::Refund,
-    B: herc20::Deploy + herc20::Fund + hbit::RedeemAsBob + herc20::Refund,
+    A: Do<hbit::CorrectlyFunded>
```
D4nte, line 27: First read: `Do` and `Execute` verbs are synonyms so it's confusing.

src/swap/do_action.rs
```diff
@@ -0,0 +1,212 @@
...
+use comit::Timestamp;
+
+#[async_trait::async_trait]
+pub trait Do<E>
+where
+    Self: CheckMemory<E> + ShouldAbort + Execute<E> + Remember<E>,
+    E: Clone + Send + Sync + 'static,
+    <Self as Execute<E>>::Args: Send + Sync,
+{
+    async fn r#do(
```
D4nte, line 11: what is that? is it because `do` is a Rust language keyword?

src/swap/do_action.rs
```diff
@@ -0,0 +1,212 @@
...
+}
+
+#[async_trait::async_trait]
+pub trait Execute<E> {
+    type Args;
+    async fn execute(&self, args: Self::Args) -> anyhow::Result<E>;
+}
+
+#[async_trait::async_trait]
+pub trait Remember<E> {
```
D4nte, line 80: Can you please comment the expected functionality brought by the various traits `Remember`, `Execute`, `CheckMemory`, `Do`?

src/swap/do_action.rs
```diff
@@ -0,0 +1,212 @@
...
+    struct FakeDatabase {
+        events: Arc<RwLock<HashMap<SwapId, ArbitraryEvent>>>,
+    }
+
+    struct FakeWallet {
+        node: Arc<RwLock<FakeBlockchain>>,
+    }
+
+    #[derive(Default)]
+    struct FakeBlockchain {
```
D4nte, line 131:
```suggestion
    struct MockBlockchain {
```

src/swap.rs
luckysori: Yeah, I haven't found good names for this yet ;(

src/swap/do_action.rs
luckysori: Yeah. I can move away from this name anyway, it's not final.

src/swap/do_action.rs
luckysori: I even googled it, but Google is a liar.

src/swap/do_action.rs
luckysori: Settled on `DoItOnce` as you suggested; coming with next PR.

src/swap/do_action.rs
D4nte: You thumbed up but did not do it?

luckysori: I _will_ do it.

PR #36, opened 2020-07-06
State clearly what is todo for config and what is unimplemented

da-kami:
Quick patch so it is easier to see what we expect from the config.


da-kami:
Config PR already on the way.

PR #38, opened 2020-07-06
Maker rate update logic

da-kami:
First commit is the (simple) rate update logic. 
The second commit shows an approach to deal with failed rate updates by comparing against the timestamp of the MidMarketRate. 
Since this is unnecessary complicated I opted for changing the `mid_market_rate` of `Maker` to be an `Option` in the third commit. When we cannot find a rate we set it to `None`. 
The `timestamp` of `MidMarketRate` was removed as we don't need it for this logic. Note that I still kept the `MidMarketRate` struct, but it is now actually just a wrapper for the inner `Rate`. We could remove it.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,6 @@
 #![allow(unreachable_code, unused_variables, clippy::unit_arg)]
 
+use nectar::maker::RateUpdateDecision;
 use nectar::{
     bitcoin, bitcoin_wallet, dai, ethereum_wallet,
     maker::TakeRequestDecision,
@@ -60,13 +61,14 @@ async fn init_maker(
 
 #[tokio::main]
 async fn main() {
-    // TODO: from config or hardcoded
-    let dai_contract_addr: comit::ethereum::Address = unimplemented!();
-
-    // TODO: Proper wallet initialisation from config
-    let bitcoin_wallet =
-        bitcoin_wallet::Wallet::new(unimplemented!(), unimplemented!(), unimplemented!()).unwrap();
-    let ethereum_wallet = ethereum_wallet::Wallet::new(unimplemented!(), unimplemented!()).unwrap();
+    let bitcoin_wallet = bitcoin_wallet::Wallet::new(
+        todo!("from config"),
+        todo!("from config"),
+        todo!("from config"),
+    )
+    .unwrap();
+    let ethereum_wallet =
+        ethereum_wallet::Wallet::new(todo!("from config"), todo!("from config")).unwrap();
 
     let maker = init_maker(bitcoin_wallet, ethereum_wallet).await;
 
@@ -90,7 +92,20 @@ async fn main() {
     loop {
         let rate = get_btc_dai_mid_market_rate().await;
         match rate {
-            Ok(new_rate) => maker.update_rate(unimplemented!()),
+            Ok(new_rate) => {
+                let reaction = maker.update_rate(new_rate); // maker should record timestamp of this
+                match reaction {
+                    Ok(RateUpdateDecision::RateUpdatedWithNewValue {
+                        next_sell_order,
+                        next_buy_order,
+                    }) => {
+                        swarm.orderbook.publish(next_sell_order.into());
+                        swarm.orderbook.publish(next_buy_order.into());
+                    }
+                    Ok(RateUpdateDecision::RateUpdated) => (),
+                    Err(err) => tracing::warn!("Rate update yielded error: {}", err),
+                }
+            }
             Err(e) => maker.track_failed_rate(e),
         }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -26,6 +26,7 @@ async fn init_maker(
     let initial_rate = get_btc_dai_mid_market_rate().await;
 
     let spread: Spread = todo!("from config");
+    let max_rate_ticker_interval_secs = todo!("from config");
 
     match (
         initial_btc_balance,
@@ -53,6 +54,7 @@ async fn init_maker(
             dai_max_sell,
             initial_rate,
             spread,
+            max_rate_ticker_interval_secs
         ),
         // TODO better error handling
         _ => panic!("Maker initialisation failed!"),
@@ -106,7 +108,7 @@ async fn main() {
                     Err(err) => tracing::warn!("Rate update yielded error: {}", err),
                 }
             }
-            Err(e) => maker.track_failed_rate(e),
+            Err(err) => tracing::warn!("Fetching rate yielded error: {}", err),
         }
 
         let bitcoin_balance = bitcoin_wallet.balance().await;
--- a/src/main.rs
+++ b/src/main.rs
@@ -107,7 +107,7 @@ async fn main() {
                 }
             }
             Err(e) => {
-                maker.invlidate_rate();
+                maker.invalidate_rate();
                 tracing::error!(
                     "Unable to fetch latest rate! Fetching rate yielded error: {}",
                     e
--- a/src/main.rs
+++ b/src/main.rs
@@ -150,7 +150,7 @@ async fn main() {
                                 swarm.orderbook.take(order);
                                 orderbook.publish(next_order.into());
                             }
-                            Ok(TakeRequestDecision::RateSucks)
+                            Ok(TakeRequestDecision::RateNotProfitable)
                             | Ok(TakeRequestDecision::InsufficientFunds)
                             | Ok(TakeRequestDecision::CannotTradeWithTaker) => {
                                 swarm.orderbook.ignore(order);
--- a/src/main.rs
+++ b/src/main.rs
@@ -26,7 +26,6 @@ async fn init_maker(
     let initial_rate = get_btc_dai_mid_market_rate().await;
 
     let spread: Spread = todo!("from config");
-    let max_rate_ticker_interval_secs = todo!("from config");
 
     match (
         initial_btc_balance,
@@ -54,7 +53,6 @@ async fn init_maker(
             dai_max_sell,
             initial_rate,
             spread,
-            max_rate_ticker_interval_secs
         ),
         // TODO better error handling
         _ => panic!("Maker initialisation failed!"),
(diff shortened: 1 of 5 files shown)


src/maker.rs
```diff
@@ -391,4 +417,44 @@ mod tests {
 
         assert_eq!(result, TakeRequestDecision::RateSucks);
     }
+
+    #[test]
+    fn rate_updated_if_rate_update_with_same_value() {
+        let mut maker = Maker {
+            mid_market_rate: MidMarketRate {
+                value: Rate::try_from(1.0).unwrap(),
+                ..Default::default()
```
luckysori: :wrench: It would be better to instantiate the current mid market rate and the new mid market rate with specific values, so that you can be sure that the test is sound. Otherwise you're relying on now to produce two different values when called consecutively, which may vary between test executions.

src/maker.rs
```diff
@@ -57,7 +57,23 @@ impl Maker {
         }
     }
 
-    pub fn update_rate(&mut self, mid_market_rate: ()) {}
+    pub fn update_rate(
+        &mut self,
+        mid_market_rate: MidMarketRate,
```
luckysori, line 62: :question: Should we check that this rate is newer than the one we have? I don't think this function can trust that the argument is the latest mid market rate, as it doesn't know where it comes from.

src/maker.rs
```diff
@@ -55,28 +51,26 @@ impl Maker {
...
     pub fn update_rate(
         &mut self,
         mid_market_rate: MidMarketRate,
     ) -> anyhow::Result<RateUpdateDecision> {
-        // if rate did not change don't create new orders
-        if self.mid_market_rate.value == mid_market_rate.value {
-            // if value did not change update to have newest timestamp
-            self.mid_market_rate = mid_market_rate;
-            return Ok(RateUpdateDecision::RateUpdated);
+        if let Some(previous_mid_market_rate) = self.mid_market_rate {
```
luckysori, line 64: :upside_down_face: It may be clearer if you have a single match statement and use match guards.

src/maker.rs
```diff
@@ -88,44 +82,38 @@ impl Maker {
...
 
-    pub fn track_failed_balance_update(&mut self, error: anyhow::Error) {}
-
-    fn rate_is_up_to_date(&self) -> bool {
-        let current_timestamp = Utc::now().timestamp();
-        let mid_market_rate_timestamp = self.mid_market_rate.timestamp.timestamp();
-        let acceptable_timestamp = (current_timestamp - self.max_rate_ticker_interval_secs);
-
-        mid_market_rate_timestamp >= acceptable_timestamp
+    pub fn track_failed_rate_update(&mut self) {
```
luckysori, line 85: :wrench: I think it would be clearer if the name of the function said what it does rather than why it would be called. You could even just inline it.

src/maker.rs
```diff
@@ -88,44 +82,38 @@ impl Maker {
...
+        self.mid_market_rate = None;
     }
 
+    pub fn track_failed_balance_update(&mut self, error: anyhow::Error) {}
```
luckysori, line 89: :wrench: You can remove it if it is unused (for now).

src/maker.rs
```diff
@@ -207,13 +200,17 @@ pub enum TakeRequestDecision {
...
+        new_sell_order: BtcDaiOrder,
+        new_buy_order: BtcDaiOrder,
     },
-    RateUpdated,
+    NoRateChange,
 }
 
+#[derive(Debug, Copy, Clone, thiserror::Error)]
+#[error("Rate not available when trying to create new {0} order.")]
+pub struct RateNotAvailable(Position);
```
luckysori, line 212: :wrench: I don't think the position adds any useful context.

src/main.rs
```diff
@@ -148,9 +152,12 @@ async fn main() {
...
                             Ok(TakeRequestDecision::RateSucks)
                             | Ok(TakeRequestDecision::InsufficientFunds)
-                            | Ok(TakeRequestDecision::CannotTradeWithTaker)
-                            | Err(_) => {
+                            | Ok(TakeRequestDecision::CannotTradeWithTaker) => {
+                                swarm.orderbook.ignore(order);
+                            }
+                            Err(e) => {
                                 swarm.orderbook.ignore(order);
+                                tracing::error!("Processing rate take request yielded error: {}", e)
```
luckysori, line 160: :wrench: 
```suggestion
                                tracing::error!("Processing taken order yielded error: {}", e)
```

src/maker.rs
```diff
@@ -136,9 +124,14 @@ impl Maker {
             return Ok(TakeRequestDecision::CannotTradeWithTaker);
         }
 
+        if self.mid_market_rate.is_none() {
+            anyhow::bail!(RateNotAvailable(order.inner.position))
+        }
+
+        let current_mid_market_rate = self.mid_market_rate.unwrap();
```
luckysori, line 131: :wrench: If you use if you match or use `unwrap_or_else` above you can avoid this unwrap.

src/mid_market_rate.rs
```diff
@@ -1,11 +1,9 @@
 use crate::Rate;
-use chrono::{DateTime, Utc};
 use std::convert::TryInto;
 
 #[derive(Debug, Copy, Clone, PartialEq)]
 pub struct MidMarketRate {
```
luckysori, line 5: :upside_down_face: Maybe it should be a tuple struct then.

src/maker.rs
```diff
@@ -441,40 +453,36 @@ mod tests {
     #[test]
     fn rate_updated_if_rate_update_with_same_value() {
```
luckysori, line 454: :mega: The name of the test has to change given the changes to the test itself.

src/maker.rs
```diff
@@ -441,40 +453,36 @@ mod tests {
...
         let new_mid_market_rate = MidMarketRate {
             value: Rate::try_from(1.0).unwrap(),
-            timestamp: Utc::now(),
         };
 
         let reaction = maker.update_rate(new_mid_market_rate).unwrap();
-        assert_eq!(reaction, RateUpdateDecision::RateUpdated);
-        assert_eq!(maker.mid_market_rate, new_mid_market_rate)
+        assert_eq!(reaction, RateUpdateDecision::NoRateChange);
+        assert_eq!(maker.mid_market_rate.unwrap(), new_mid_market_rate)
```
luckysori, line 468: :wrench: You can use `matches!` to avoid this unwrap.

src/maker.rs
```diff
@@ -441,40 +453,36 @@ mod tests {
...
         let new_mid_market_rate = MidMarketRate {
             value: Rate::try_from(2.0).unwrap(),
-            timestamp: Utc::now(),
         };
 
         let reaction = maker.update_rate(new_mid_market_rate).unwrap();
-        assert!(matches!(reaction, RateUpdateDecision::RateUpdatedWithNewValue {..}));
-        assert_eq!(maker.mid_market_rate, new_mid_market_rate)
+        assert!(matches!(reaction, RateUpdateDecision::RateChange {..}));
+        assert_eq!(maker.mid_market_rate.unwrap(), new_mid_market_rate)
```
luckysori, line 486: :wrench: You can use `matches!` to avoid this unwrap.

src/maker.rs
da-kami: I think they are, this might be from an outdated version.

src/maker.rs
da-kami: This check would not have much value given the current implementation - and the Kraken API. The ticker actually does not give us a timestamp. We could give it a timestamp inside `get_btc_dai_mid_market_rate` but there is not much value in that. 
Thus I removed the complete timestamp check for now. Whenever this function is called the maker _always_ assumes it is a new one.

src/maker.rs
da-kami: I did it like this because it is an early return for a particular `Some` case - the other `Some` case will be treated like the `None` case.
I think if I don't do it like this I would have to duplicate the logic after the `if let` - but maybe I'm missing something :)

src/maker.rs
da-kami: I kindoff like the function better than just making the `mid_market_rate` public. This way it is a clearer interface I feel. 
I renamed it to `invalidate_rate()`.

src/maker.rs
da-kami: That has been in there for long already :D 
Removed!

src/maker.rs
da-kami: I thought it might add value if this ever happens in `process_taken_order` - don't have strong feelings about it though.

src/maker.rs
da-kami: This might do: 

```
assert_eq!(maker.mid_market_rate, Some(new_mid_market_rate))
```

Matches is probably an overkill here :)

src/maker.rs
luckysori: True, this was a comment on the first commit.

src/maker.rs
luckysori: I have pushed https://github.com/coblox/nectar/pull/38/commits/f9eab8fec9af4d3599d1641ae7d8a667f2e34ded onto this branch to show you what I meant. See if you like it :D

src/maker.rs
luckysori: Makes sense :D I pushed https://github.com/coblox/nectar/pull/38/commits/a14ff830809521c4384db57ff6286a71fc110f35 to fix the typo in the name.

src/maker.rs
```diff
@@ -51,13 +51,34 @@ impl Maker {
...
             spread,
             ongoing_takers: Default::default(),
         }
     }
 
-    pub fn update_rate(&mut self, mid_market_rate: ()) {}
+    pub fn update_rate(
+        &mut self,
+        mid_market_rate: MidMarketRate,
+    ) -> anyhow::Result<RateUpdateDecision> {
```
D4nte, line 63:
```
	) -> anyhow::Result<Option((Order, Order))> {
```
I don't think the `RateUpdateDecision` enum adds much value. It's unneeded complexity, just return an option and if the option contain orders, publish them.

src/main.rs
```diff
@@ -131,9 +152,12 @@ async fn main() {
                             }
                             Ok(TakeRequestDecision::RateSucks)
```
D4nte, line 153: As previously stated, if you treat of these enum variants (and the Err case) the same way then it's a good indicator you don't need them.

src/maker.rs
```diff
@@ -61,17 +61,19 @@ impl Maker {
         &mut self,
         mid_market_rate: MidMarketRate,
     ) -> anyhow::Result<RateUpdateDecision> {
-        if let Some(previous_mid_market_rate) = self.mid_market_rate {
-            if previous_mid_market_rate == mid_market_rate {
-                return Ok(RateUpdateDecision::NoRateChange);
+        match self.mid_market_rate {
+            Some(previous_mid_market_rate) if previous_mid_market_rate == mid_market_rate => {
```
da-kami, line 65: I did not know one can add to the match expression like this, neat!

src/maker.rs
da-kami: I only kept that enum because an option of a tuple struct - `Option<(BtcDaiOrder, BtcDaiOrder)>` -would imply the sell and buy order arbitrarily and I did not like that.

That can also be solved by wrapping the orders by another struct - I pushed this as a change for now, but I feel the complexity is more or less the same. 

Reducing the output to just an `Option` would imply that we don't bubble up the `Result` of `new_xxx_order()`.

src/main.rs
da-kami: I agree, that one should not introduce code before needing/using it, but I feel if I don't use these enum variants I would have to model different errors. But I did not want to treat these scenarios as errors.
I think the enum variants are nice because they sum up the expected outputs - and they helped. me when writing tests. 

The only unnecessary complexity I see is when treating the variants in `main`.

D4nte: In this case, please use better language than `sucks`.

PR #39, opened 2020-07-07
Support Herc20<->Hbit

luckysori:
I haven't added another integration test because the function is easy to review and all the individual pieces are tested. We can add the integration test (or even a mock test) if we want to, but to do it properly we would have to clean up the integration test that already exists first. I think we can do that in a separate PR.

### DIFF
--- a/src/swap.rs
+++ b/src/swap.rs
@@ -14,7 +14,7 @@ use futures::future;
 
 pub use alice::WatchOnlyAlice;
 pub use bob::WalletBob;
-pub use do_action::{BetaLedgerTime, Do, Execute, Next};
+pub use do_action::{AlphaLedgerTime, BetaExpiry, BetaLedgerTime, Do, Execute, Next};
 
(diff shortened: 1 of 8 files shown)


src/swap.rs
```diff
@@ -14,96 +14,146 @@ use futures::future;
...
+        Ok(future::Either::Left((_hbit_redeemed, _))) => Ok(()),
+        Ok(future::Either::Right((_hbit_refunded, _))) => Ok(()),
+        Err(either) => {
+            let (error, _other_future) = either.factor_first();
+            Err(error)
+        }
+    }
+}
+
+/// Execute a Hbit<->Herc20 swap.
```
D4nte, line 92:
```suggestion
/// Execute a Herc20<->Hbit swap.
```

luckysori:
> I know the `CheckMemory` trait was not introduced here but.. what is it about? Can you clarify?

Yep, yep. It's in my TODOs.

PR #40, opened 2020-07-08
Event loop optimisation and unreserving funds logic

da-kami:
Free funds and "remove taker from ongoing takers" logic.

This includes an update for the event loop model that allows us to process network events, swap execution events and balance/rate updates separately using `futures::select!`.

I would try to get this merged so that the new structure is in. Want to avoid rebasing this too often 😬

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,20 +1,32 @@
 #![allow(unreachable_code, unused_variables, clippy::unit_arg)]
+#![recursion_limit = "256"]
 
 use anyhow::Context;
+use comit::{hbit, herc20};
+use futures::channel::mpsc::{Receiver, Sender};
+use futures::FutureExt;
+use futures::StreamExt;
+use futures::{Future, SinkExt};
+use futures_timer::Delay;
 use nectar::config::settings;
+use nectar::maker::Free;
 use nectar::maker::PublishOrders;
+use nectar::network::Taker;
+use nectar::order::Position;
 use nectar::{
-    bitcoin_wallet, config,
+    bitcoin, bitcoin_wallet, config,
     config::Settings,
-    ethereum_wallet,
+    dai, ethereum_wallet,
     maker::TakeRequestDecision,
     mid_market_rate::get_btc_dai_mid_market_rate,
     network::{self, Nectar, Orderbook},
     options::{self, Options},
-    Maker, Spread,
+    Maker, MidMarketRate, Spread,
 };
 use std::time::Duration;
 use structopt::StructOpt;
+
+const ENSURED_CONSUME_ZERO_BUFFER: usize = 0;
 
 async fn init_maker(
     bitcoin_wallet: bitcoin_wallet::Wallet,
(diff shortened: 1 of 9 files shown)


src/order.rs
```diff
@@ -25,6 +25,10 @@ impl BtcDaiOrder {
         mid_market_rate: Rate,
         spread: Spread,
     ) -> anyhow::Result<BtcDaiOrder> {
+        if base_balance < base_reserved_funds + base_fees {
```
da-kami, line 28: I added these checks because if we calculate a minus number the calculation panics. 

There are other solutions for this - instead of using the `-` operator we could also go for the underlying functions that return Options. But I thought avoiding the error is probably the more intuitive way.
Since we use this function on multiple occasions outside I decided that the check has to more in here.

src/main.rs
```diff
@@ -63,6 +74,216 @@ async fn init_maker(
...
+) -> (
+    Pin<Box<dyn Future<Output = comit::Never> + Send>>,
+    Receiver<anyhow::Result<bitcoin::Amount>>,
+) {
+    let (mut sender, receiver) =
+        futures::channel::mpsc::channel::<anyhow::Result<bitcoin::Amount>>(5);
+
+    let future = async move {
+        loop {
+            let balance = wallet.balance().await;
```
luckysori, line 117: :thinking: We could keep track of the last value we sent through the channel and only send again if the new value we get from the wallet is different.

luckysori: This would apply to all of these updates.

src/main.rs
```diff
@@ -63,6 +74,216 @@ async fn init_maker(
...
+                    }
+                }
+                Ok(TakeRequestDecision::RateNotProfitable)
+                | Ok(TakeRequestDecision::InsufficientFunds)
+                | Ok(TakeRequestDecision::CannotTradeWithTaker) => {
+                    swarm.orderbook.ignore(order);
+                }
+                Err(e) => {
+                    swarm.orderbook.ignore(order);
+                    tracing::error!("Processing rate take request yielded error: {}", e)
```
luckysori, line 217:
```suggestion
                    tracing::error!("Processing taken order yielded error: {}", e)
```

src/maker.rs
```diff
@@ -468,4 +551,35 @@ mod tests {
...
+            btc_fee: bitcoin::Amount::from_btc(0.1).unwrap(),
+            dai_fee: dai::Amount::from_dai_trunc(0.1).unwrap(),
+            ..Default::default()
+        };
+
+        let taker = Taker { 0: 0 };
+
+        maker.ongoing_takers.insert(taker).unwrap();
+
+        let unreserve_btc = Free::Btc(bitcoin::Amount::from_btc(0.5).unwrap());
```
luckysori, line 569: :wrench:
```suggestion
        let free_btc = Free::Btc(bitcoin::Amount::from_btc(0.5).unwrap());
```

luckysori: And others.

src/maker.rs
```diff
@@ -468,4 +551,35 @@ mod tests {
...
+            btc_reserved_funds: bitcoin::Amount::from_btc(1.1).unwrap(),
+            dai_reserved_funds: dai::Amount::from_dai_trunc(1.1).unwrap(),
+            btc_fee: bitcoin::Amount::from_btc(0.1).unwrap(),
+            dai_fee: dai::Amount::from_dai_trunc(0.1).unwrap(),
+            ..Default::default()
+        };
+
+        let taker = Taker { 0: 0 };
+
+        maker.ongoing_takers.insert(taker).unwrap();
```
luckysori, line 567: :upside_down_face: Perhaps it would be better to `process_taken_order` as part of the test instead of building the state manually. Or not. Just an idea.

src/main.rs
da-kami: We can, but that means we store state twice, once in the main loop and once in the Maker.

src/main.rs
```diff
@@ -63,6 +74,216 @@ async fn init_maker(
     }
 }
 
+// TODO: consider introducing components for this (also balance updates below),
+//  Rough idea: Extend MidMarketRate (or introduce new struct) to give us an async next function, hide the fact that it is using channels behind the next function of the component.
```
D4nte, line 78: I don't think `next` is the appropriate terminology, it is not an iterator.

src/main.rs
D4nte: I think we should let the maker decide whether the balance/rate changed and what to do from there.

For example, a rate could cahnge within the spread and the maker could decide not to do anything about it.

src/main.rs
```diff
@@ -63,6 +74,216 @@ async fn init_maker(
...
+            }
+
+            Delay::new(Duration::from_secs(update_interval_secs)).await;
+        }
+    };
+
+    (Box::pin(future), receiver)
+}
+
+// TODO: turn this into an update of DAI and Ether and keep track of the Ether balance in the maker as well
```
D4nte, line 130: 📣 Why would you want to track the Ether balance in the maker?

src/main.rs
```diff
@@ -63,6 +74,216 @@ async fn init_maker(
     }
 }
 
+// TODO: consider introducing components for this (also balance updates below),
+//  Rough idea: Extend MidMarketRate (or introduce new struct) to give us an async next function, hide the fact that it is using channels behind the next function of the component.
+#[allow(clippy::type_complexity)]
+fn init_rate_updates(
+    update_interval_secs: u64,
```
D4nte, line 81:
```suggestion
    update_interval_secs: Duration,
```

src/main.rs
```diff
@@ -63,6 +74,216 @@ async fn init_maker(
...
+            Delay::new(Duration::from_secs(update_interval_secs)).await;
+        }
+    };
+
+    (Box::pin(future), receiver)
+}
+
+#[allow(clippy::type_complexity)]
+fn init_bitcoin_balance_updates(
+    update_interval_secs: u64,
```
D4nte, line 106:
```suggestion
    update_interval_secs: Duration,
```

src/main.rs
```diff
@@ -63,6 +74,216 @@ async fn init_maker(
...
+        }
+    };
+
+    (Box::pin(future), receiver)
+}
+
+// TODO: turn this into an update of DAI and Ether and keep track of the Ether balance in the maker as well
+#[allow(clippy::type_complexity)]
+fn init_dai_balance_updates(
+    update_interval_secs: u64,
```
D4nte, line 133:
```suggestion
    update_interval_secs: Duration,
```

src/order.rs
D4nte: Please add a proptest on the function to ensure you solved the panic.

src/main.rs
da-kami: Because it will be part of the bookkeeping - it is part of reserving funds, because it is necessary for the settlement (i.e. swap execution) and it has to be part of considering to create a new order to ensure the order can be executed.

src/maker.rs
da-kami: That's what I had in mind with crating more elaborate roundtrip tests. I can also start doing it here. I tried to only test one function per test so far so it is clear what is tested.

src/main.rs
D4nte: I don't agree on the fact that it is part of book keeping. For the sake of finishing the project we should assume there is always enough gas to execute.

src/main.rs
```diff
@@ -58,15 +68,161 @@ async fn init_maker(
...
 }
 
+fn init_rate_updates(
+    update_interval_secs: u64,
+) -> (
+    Pin<Box<dyn Future<Output = comit::Never> + Send>>,
+    Receiver<anyhow::Result<MidMarketRate>>,
+) {
+    let (mut sender, receiver) =
+        futures::channel::mpsc::channel::<anyhow::Result<MidMarketRate>>(5);
```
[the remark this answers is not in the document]
da-kami: Changing it to `0` (using a constant) , we should immediately consume on all channels due to the loop; given the current design it cannot not happen that a sender sends something while the receiver has not received the previous message.

@thomaseizinger I think this should also be true for the execution channel, because I am cloning the `sender` upon spawning the execution. Like for the other channels there should not be back-pressure, thus I think I can set all of the channels buffers to `0`. Please correct me if I am mistaken.

src/main.rs
da-kami: OK for me for now, once https://github.com/coblox/nectar/pull/42 is in this logic makes sense again. I'll keep this comment as `Unresolved` for now, to keep this in mind.

src/main.rs
da-kami: I just removed the `TODO` for now, no need to deal with this before the project finishes.

src/order.rs
da-kami: Good call! @D4nte you might want to have another look, the guarding logic changed. I wonder if the complexity can be reduced.

src/main.rs
da-kami: I removed the TODO from code, will adapt the TODO on basecamp (keep track that tracking Ether for fees is missing so we can potentially build it into a follow up).

src/main.rs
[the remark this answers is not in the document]
da-kami: Actually on second thought I am not sure this makes much sense...

I had to tell clippy to allow an unnecessary clone for this:

https://github.com/coblox/nectar/blob/b879d738cb92aa8eb4560025a730adbc16e0e48c/src/main.rs#L222-L225

I doubt that this really has the effect I think it has. @thomaseizinger if you have time it would be cool to discuss this :)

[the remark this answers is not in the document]
da-kami: Note: After discussion with @thomaseizinger we moved the clone into the loop (clippy does not complain about it there) - we looked at the expanded `select!` code because this is weird behaviour (the compiler should have complained already, because the clone is expected!). 

Thomas will look into this further and might upstream a question with a simplified example :)

src/main.rs
```diff
@@ -63,6 +75,213 @@ async fn init_maker(
...
+
+async fn execute_swap(sender: Sender<FinishedSwap>) -> anyhow::Result<()> {
+    let position: Position =
+        todo!("decision what kind of what swap it is hbit->herc20 or herc20->hbit");
+
+    let taker: Taker = todo!("Taker has to be available after execution, e.g. load from db");
+
+    match position {
+        Position::Sell => {
+            let beta_params: hbit::Params = unimplemented!();
```
thomaseizinger, line 162: This unimplemented and probably the `todo!`s above are the reason for the ownership not being enforced :)

src/main.rs
```diff
@@ -63,6 +75,213 @@ async fn init_maker(
...
+
+            // TODO: await hbit->herc20 swap execution
+
+            sender.try_send(FinishedSwap::new(
+                Free::Btc(beta_params.asset.into()),
+                taker,
+            ))?;
```
thomaseizinger, line 169: Better use `sender.send().await` here.

src/main.rs
thomaseizinger: Here is a playground of a smaller example where the ownership is enforced:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d1c2d079d84518e90121f6a3023412b1

src/main.rs
da-kami: thanks, done!

PR #43, opened 2020-07-10
Use dai contract from config in ethereum wallet

da-kami:


src/main.rs
```diff
@@ -295,8 +295,12 @@ async fn main() {
...
     )
     .unwrap();
-    let ethereum_wallet =
-        ethereum_wallet::Wallet::new(unimplemented!(), settings.ethereum.node_url).unwrap();
+    let ethereum_wallet = ethereum_wallet::Wallet::new(
+        unimplemented!(),
+        settings.ethereum.node_url,
+        dai_contract_addr,
+    )
+    .unwrap();
```
D4nte, line 303: It'd be better to use `expect()`.

da-kami: I'd say we clean that up in another iteration. 
There are several `unwraps()` to be cleaned up.

PR #44, opened 2020-07-10
Introduce seed file

D4nte:
Very similar to cnd, the secret key is now a sha-256 of the seed.
The seed is stored in a PEM file which is in the data dir as defined in the config.


src/main.rs
```diff
@@ -286,17 +286,20 @@ async fn main() {
         .and_then(Settings::from_config_file_and_defaults)
         .expect("Could not initialize configuration");
 
+    let seed = config::Seed::from_file_or_generate(settings.data.dir)
+        .expect("Could not retrieve/initialize seed")
+        .into();
+
     let dai_contract_addr: comit::ethereum::Address = settings.ethereum.dai_contract_address;
 
     // TODO: Proper wallet initialisation from config
```
da-kami, line 295:
```suggestion
```

src/config/seed.rs
```diff
@@ -0,0 +1,209 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Seed([*****])")
+    }
+}
+
+impl fmt::Display for Seed {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
```
luckysori, line 86: :question: Is this standard or should we just not implement these for `Seed`?

src/config/seed.rs
```diff
@@ -0,0 +1,209 @@
...
+use std::{
+    ffi::OsStr,
+    fmt,
+    fs::{self, File},
+    io::{self, Write},
+    path::{Path, PathBuf},
+};
+
+#[derive(Clone, Copy, PartialEq)]
+pub struct Seed(seed::Seed);
```
luckysori, line 13: :+1: Nice! At first I was confused with the 2 seeds, but wrapping the original one in a `config::Seed` for deserialization and serialization makes a lot of sense.

src/config/seed.rs
D4nte: I copied that from `comit-rs`. I quite like it because if it's not there, a developer could decided to derive `Debug` or `Display` and not think about the consequences.

src/config/seed.rs
D4nte: Yes, in `comit-rs` it's called `RootSeed` but the context is different. I think it make sense to keep the "real" `Seed` agnostic on how it's configured/written on disk/etc.

PR #46, opened 2020-07-10
Respawn unfinished swaps before entering main loop

luckysori:
Quick and dirty attempt at respawning swaps to show a possible `load_all` interface.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,28 +2,24 @@
 #![recursion_limit = "256"]
 
 use anyhow::Context;
-use comit::{hbit, herc20};
-use futures::channel::mpsc::{Receiver, Sender};
-use futures::FutureExt;
-use futures::StreamExt;
-use futures::{Future, SinkExt};
+use futures::{
+    channel::mpsc::{Receiver, Sender},
+    Future, FutureExt, SinkExt, StreamExt,
+};
 use futures_timer::Delay;
-use nectar::config::settings;
-use nectar::maker::Free;
-use nectar::maker::PublishOrders;
-use nectar::network::Taker;
-use nectar::order::Position;
 use nectar::{
     bitcoin, bitcoin_wallet, config,
-    config::Settings,
+    config::{settings, Settings},
     dai, ethereum_wallet,
-    maker::TakeRequestDecision,
+    maker::{Free, PublishOrders, TakeRequestDecision},
     mid_market_rate::get_btc_dai_mid_market_rate,
-    network::{self, Nectar, Orderbook},
+    network::{self, Nectar, Orderbook, Taker},
     options::{self, Options},
+    order::Position,
+    swap::{self, hbit, herc20, Database, SwapKind},
     Maker, MidMarketRate, Spread,
 };
-use std::time::Duration;
+use std::{sync::Arc, time::Duration};
 use structopt::StructOpt;
 
 const ENSURED_CONSUME_ZERO_BUFFER: usize = 0;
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -331,6 +327,12 @@ async fn main() {
     let (swap_execution_finished_sender, swap_execution_finished_receiver) =
         futures::channel::mpsc::channel::<FinishedSwap>(ENSURED_CONSUME_ZERO_BUFFER);
 
+    let db = Arc::new(Database::new(todo!(
+        "try to load from config, otherwise default?"
```
D4nte, line 331: 👍

src/main.rs
```diff
@@ -355,6 +357,35 @@ async fn main() {
...
 
+fn respawn_swaps(
+    db: Arc<Database>,
+    bitcoin_wallet: Arc<bitcoin_wallet::Wallet>,
+    ethereum_wallet: Arc<ethereum_wallet::Wallet>,
+    bitcoin_connector: Arc<comit::btsieve::bitcoin::BitcoindConnector>,
+    ethereum_connector: Arc<comit::btsieve::ethereum::Web3Connector>,
+    swap_execution_finished_sender: Sender<FinishedSwap>,
+) -> anyhow::Result<()> {
+    // Assumption: all swaps in the database are active swaps
```
D4nte, line 368: Hum, good one. So we need a call to delete the swap from the DB after it's done and we saved it in the csv file. Can you add this call to this PR?

da-kami, line 368: Not sure if you already have this in mind, thought I better mention it. 

I think when we re-spawn we will have to do bookkeeping as well, i.e. 

- load swap, evaluate if unfinished
- if unfinished reserve funds in maker
- re-spawn execution with `Sender<FinishedSwap>`
- (free funds logic will be called eventually)

Otherwise we might run into the problem, that we try to free more than is reserved.

src/main.rs
```diff
@@ -355,6 +357,35 @@ async fn main() {
...
+
+    for swap in db.load_all()?.into_iter() {
```
D4nte, line 370: 👍

src/main.rs
luckysori: Done with https://github.com/coblox/nectar/pull/46/commits/9d761589ba13e0f001f74ffbad5edec7650d5678.

src/main.rs
```diff
@@ -166,6 +164,7 @@ async fn execute_swap(sender: Sender<FinishedSwap>) -> anyhow::Result<()> {
 
             if let Err(e) = sender
                 .send(FinishedSwap::new(
+                    swap_id,
                     Free::Dai(beta_params.asset.into()),
```
D4nte, line 168: Hum, what is that? 🤔

src/main.rs
```diff
@@ -264,14 +263,31 @@ fn handle_dai_balance_update(
     unimplemented!()
 }
 
+// TODO: I don't think `finished_swap` should be an Option
+fn handle_finished_swap(finished_swap: Option<FinishedSwap>, maker: &mut Maker, db: &Database) {
+    if let Some(finished_swap) = finished_swap {
+        maker.process_finished_swap(finished_swap.funds_to_free, finished_swap.taker);
+
```
D4nte, line 270: Add todo "save in csv"

src/main.rs
luckysori: I did find the name a bit confusing if that's what you mean.

luckysori: I think for the purposes of writing it to the history file you will need to modify this.

src/main.rs
luckysori: Done!

src/main.rs
luckysori: Added as a TODO.

luckysori:
Closing in favour of #49.

PR #47, opened 2020-07-10
Balance update logic

da-kami:
First Commit: More complex logic, that takes the max-amount into account.
After sleeping over this on the weekend I feel it is unnecessarily complex at the moment.
I think it is totally fine to publish orders upon every balance change, even if that means "spamming" the network if the order's amounts did not change.

### DIFF
--- a/src/dai.rs
+++ b/src/dai.rs
@@ -162,10 +162,6 @@ impl Amount {
         Ok(bitcoin::Amount::from_sat(sats))
     }
 
-    pub fn symbol(self) -> String {
-        "DAI".to_owned()
-    }
-
     pub fn checked_add(self, rhs: Amount) -> Option<Amount> {
         self.0.checked_add(&rhs.0).map(Amount)
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,10 +71,9 @@ pub fn data_dir() -> Option<std::path::PathBuf> {
 }
 
 #[derive(Debug, Copy, Clone, strum_macros::Display)]
+#[strum(serialize_all = "UPPERCASE")]
 pub enum Symbol {
-    #[strum(serialize = "BTC")]
     Btc,
-    #[strum(serialize = "DAI")]
     Dai,
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,5 +70,27 @@ pub fn data_dir() -> Option<std::path::PathBuf> {
         .map(|proj_dirs| proj_dirs.data_dir().to_path_buf())
 }
 
+#[derive(Debug, Copy, Clone, strum_macros::Display)]
+pub enum Symbol {
+    #[strum(serialize = "BTC")]
+    Btc,
+    #[strum(serialize = "DAI")]
+    Dai,
+}
+
 #[cfg(all(test, feature = "test-docker"))]
 pub mod test_harness;
+
+#[cfg(test)]
+mod tests {
+    use crate::Symbol;
+
+    #[test]
+    fn symbol_serializes_correctly() {
+        let btc = Symbol::Btc;
+        let dai = Symbol::Dai;
+
+        assert_eq!(String::from("BTC"), btc.to_string());
+        assert_eq!(String::from("DAI"), dai.to_string());
+    }
+}
--- a/src/main.rs
+++ b/src/main.rs
@@ -225,8 +225,8 @@ fn handle_rate_update(
 ) {
     match rate_update {
         Ok(new_rate) => {
-            let reaction = maker.update_rate(new_rate);
-            match reaction {
+            let result = maker.update_rate(new_rate);
+            match result {
                 Ok(Some(PublishOrders {
                     new_sell_order,
                     new_buy_order,
@@ -253,7 +253,22 @@ fn handle_btc_balance_update(
     maker: &mut Maker,
     swarm: &mut libp2p::Swarm<Nectar>,
 ) {
-    unimplemented!()
+    match btc_balance_update {
+        Ok(btc_balance) => match maker.update_bitcoin_balance(btc_balance) {
+            Ok(Some(new_sell_order)) => {
+                swarm.orderbook.publish(new_sell_order.into());
+            }
+            Ok(None) => (),
+            Err(e) => tracing::warn!("Bitcoin balance update yielded error: {}", e),
+        },
+        Err(e) => {
+            maker.invalidate_bitcoin_balance();
+            tracing::error!(
+                "Unable to fetch bitcoin balance! Fetching balance yielded error: {}",
+                e
+            );
+        }
+    }
 }
 
 fn handle_dai_balance_update(
@@ -261,7 +276,22 @@ fn handle_dai_balance_update(
     maker: &mut Maker,
     swarm: &mut libp2p::Swarm<Nectar>,
 ) {
-    unimplemented!()
+    match dai_balance_update {
+        Ok(dai_balance) => match maker.update_dai_balance(dai_balance) {
+            Ok(Some(new_buy_order)) => {
+                swarm.orderbook.publish(new_buy_order.into());
+            }
+            Ok(None) => (),
+            Err(e) => tracing::warn!("Dai balance update yielded error: {}", e),
+        },
+        Err(e) => {
+            maker.invalidate_dai_balance();
+            tracing::error!(
+                "Unable to fetch dai balance! Fetching balance yielded error: {}",
+                e
+            );
+        }
+    }
 }
 
 struct FinishedSwap {
--- a/src/maker.rs
+++ b/src/maker.rs
@@ -6,9 +6,8 @@ use crate::{
     network::Order,
     order::{BtcDaiOrder, Position},
     rate::Spread,
-    MidMarketRate, PeersWithOngoingTrades, Rate, Symbol,
+    MidMarketRate, PeersWithOngoingTrades, Symbol,
 };
-use std::convert::TryFrom;
 
 #[derive(Debug, PartialEq)]
 pub enum NewOrder {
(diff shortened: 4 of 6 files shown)


src/dai.rs
```diff
@@ -163,7 +164,7 @@ impl Amount {
     }
 
     pub fn symbol(self) -> String {
```
D4nte, line 166: You don't need to pass `self` here. Also, why not using the constant directly? (especially that you made the const `pub`).

src/main.rs
```diff
@@ -253,15 +253,51 @@ fn handle_btc_balance_update(
     maker: &mut Maker,
     swarm: &mut libp2p::Swarm<Nectar>,
 ) {
-    unimplemented!()
+    match btc_balance_update {
+        Ok(btc_balance) => {
+            let reaction = maker.update_bitcoin_balance(btc_balance);
```
D4nte, line 258: This `reaction` terminology is uncommon. Just name it from its type: `result`.

src/maker.rs
```diff
@@ -78,38 +78,100 @@ impl Maker {
...
+                return Ok(None);
+            }
+        }
+
+        // TODO: Evaluate if we even care... we could also just "spam" with creating new orders and simplify this logic (add tests once we decide)
+
+        // we know that the balance changed and have to take in account the max_sell_amount
+        match (self.btc_max_sell_amount, self.btc_balance) {
+            (Some(max_sell_amount), Some(previous_balance))
+                if balance >= max_sell_amount && !(previous_balance < max_sell_amount) =>
```
D4nte: We are duplicating the logic of taking in account the `max_sell_amount` (the logic is also present when creating a new order).

What if instead, we just create a new order and compare it to the last order created? If it's the same amount (because the max_sell_amount was capping the order anyway and the rate did not change) then we returned `None`, other returned the order created.

src/maker.rs
```diff
@@ -141,45 +173,55 @@ impl Maker {
...
-                            return Ok(TakeRequestDecision::RateNotProfitable);
-                        }
-
-                        let updated_dai_reserved_funds =
-                            self.dai_reserved_funds.clone() + order.quote;
-                        if updated_dai_reserved_funds > self.dai_balance {
-                            return Ok(TakeRequestDecision::InsufficientFunds);
+                        match self.dai_balance.clone() {
+                            Some(dai_balance) => {
+                                // We are buying BTC for DAI
```
D4nte, line 178:
```suggestion
                        match self.dai_balance {
                            Some(ref dai_balance) => {
                                // We are buying BTC for DAI
```
Clone later :)

src/maker.rs
```diff
@@ -141,45 +173,55 @@ impl Maker {
...
+                                // Given an order rate of: 1:9000
+                                // It is NOT profitable to buy, if the current rate is greater than the order rate.
+                                // 1:8800 -> We give less DAI for getting BTC -> Good.
+                                // 1:9200 -> We have to give more DAI for getting BTC -> Sucks.
+                                if order_rate > current_profitable_rate {
```
D4nte, line 183:
```suggestion
                                if order_rate >= current_profitable_rate {
```
Otherwise it means we never accept orders we just created!

src/maker.rs
```diff
@@ -231,6 +273,10 @@ pub struct PublishOrders {
 #[error("Rate not available when trying to create new {0} order.")]
 pub struct RateNotAvailable(Position);
 
+#[derive(Debug, Clone, thiserror::Error)]
+#[error("{0} balance not available.")]
+pub struct BalanceNotAvailable(String);
```
D4nte, line 278:
```suggestion
pub struct BalanceNotAvailable(Symbol);
```
Please use a type.

src/maker.rs
```diff
@@ -279,24 +349,21 @@ mod tests {
...
-        )
+        assert_eq!(maker.btc_reserved_funds, btc(1.5))
     }
 
     #[test]
     fn btc_funds_reserved_upon_taking_sell_order_with_fee() {
         let mut maker = Maker {
-            btc_balance: bitcoin::Amount::from_btc(3.0).unwrap(),
-            btc_fee: bitcoin::Amount::from_btc(1.0).unwrap(),
+            btc_balance: some_btc(3.0),
```
D4nte, line 358:
```suggestion
            btc_balance: Some(btc(3.0)),
```

I am not sure these helpers functions are that helpful, you save 1 character per call.

src/maker.rs
D4nte: Please add a test for this logic: take an order we just created.

src/dai.rs
da-kami: I remove the function for now and just use the `const` exposed by the assets, can potentially use `enum Symbol` introduced by https://github.com/coblox/nectar/pull/48/files

src/maker.rs
da-kami: Reduced the logic to always publishing an order if the balance' value changed. 
Recorded possible follow-up item for this.

src/maker.rs
da-kami: This is the **exit** path, it should be fine as is. 

I added one test for buy and one test for sell so it is more obvious:

```published_sell_order_can_be_taken```
```published_buy_order_can_be_taken```

File: src/maker.rs (hunk 7 above)
Comment (da-kami) [reply src/maker.rs:358]: It's just faster when typing because I only have to auto-complete one function call :D 
If you feel strong about that one I'd suggest a follow up.

File: src/maker.rs (hunk 4 above)
Comment (da-kami) [reply src/maker.rs:178]: I did the `ref` and `*` (deref) upon comparison below. Is that more idiomatic than the clone?

File: src/maker.rs (hunk 6 above)
Comment (da-kami) [reply src/maker.rs:278]: I think we could reuse the symbol introduced by https://github.com/coblox/nectar/pull/48/files here? 
I kept it as string for now, because I treat it as a `const` of the module atm. @D4nte would you mind creating a follow up for this?

File: src/maker.rs [hunk 8]
```diff
@@ -141,45 +173,55 @@ impl Maker {
...
-                        if updated_btc_reserved_funds > self.btc_balance {
-                            return Ok(TakeRequestDecision::InsufficientFunds);
+                        match self.btc_balance {
+                            Some(btc_balance) => {
+                                // We are selling BTC for DAI
+                                // Given an order rate of: 1:9000
+                                // It is NOT profitable to sell, if the current rate is smaller than the order rate.
+                                // 1:8800 -> We get less DAI for our BTC -> Sucks.
+                                // 1:9200 -> We get more DAI for our BTC -> Good.
+                                if order_rate < current_profitable_rate {
```
luckysori, line 211: :upside_down_face: We could write an `is_as_good_as(rate)` method on `BtcDaiOrder` which hides this complexity and makes it more readable.

src/maker.rs
D4nte: Hum no, I would not re-use `history::Symbol`. Nothing in `history` is to be exposed. Only things that are exposed in `history` are:

- `History::new`
- `History::write`
- `impl From<..::Swap< for history::Swap` (to be done)

src/maker.rs
D4nte: The `ref` way is not more idiomatic, it is just better because you avoid a `clone`. You use less memory and it's faster because you are not copying data.

da-kami:
> I don't think we are actually reserving funds when creating new orders after a balance update. Reserving funds seems to be tied to `process_taken_order`, but it should probably be done with `new_buy_order` and `new_sell_order`. It's better to make those operations atomic anyway.

Reserving funds is currently tied to a taker wanting to trade with us. It has nothing to do with the order, but with the execution of the order. We reserve when we go for execution (with the assumption that it will be settled) and free once it was actually settled.
 
Reserving the money when creating the order opens up a bigger problem space, because we would have to distinguish freeing because an order becomes outdated and freeing after a swap.

Could you elaborate why think it is necessary to reserve upon creating an order?

luckysori:
> Reserving funds is currently tied to a taker wanting to trade with us. It has nothing to do with the order, but with the execution of the order. We reserve when we go for execution (with the assumption that it will be settled) and free once it was actually settled (be it that the other party got it or we got it back).

Yep, you are right. I got confused because in `handle_network_event`, after processing a taken order we both confirm it with the orderbook _and_ publish a new one. Given that, I wrongly thought that we were reserving funds for the one to be published.

My bad. Thanks for the clarification :relaxed:

src/dai.rs
da-kami: introduced enum

src/maker.rs
da-kami: introduced separate enum

src/maker.rs
D4nte: > This is the **exit** path, it should be fine as is.

I am not sure to understand this comment. The code seems wrong to me and no test seems to have caught that. Why is it fine?

da-kami: The code is not wrong. We exit early if the rate is not profitable.
If the rate is the same it is profitable.
I added 2 tests to show that it works as expected.

src/lib.rs
```diff
@@ -70,5 +70,27 @@ pub fn data_dir() -> Option<std::path::PathBuf> {
         .map(|proj_dirs| proj_dirs.data_dir().to_path_buf())
 }
 
+#[derive(Debug, Copy, Clone, strum_macros::Display)]
```
D4nte, line 73: use `serialize_all` instead

src/maker.rs
da-kami: Soooo much love - I implemented it 😄

src/maker.rs
```diff
@@ -167,64 +166,46 @@ impl Maker {
                 let current_profitable_rate = self
                     .spread
                     .apply(current_mid_market_rate.into(), order.inner.position)?;
-                let order_rate = Rate::try_from(order.inner.clone())?;
+
+                if !order.inner.is_as_good_as(current_mid_market_rate)? {
```
D4nte, line 170: `is_as_profitable_as`. "good" is a too generic.

PR #48, opened 2020-07-10
Add CSV File history

D4nte:
Most of the heavy lifting is now done. Just need to fine tune the data and implement `Into<Trade> for FinishedSwap`.

This can be done when integrating this to the main loop.

### DIFF
--- a/src/history.rs
+++ b/src/history.rs
@@ -1,0 +1,295 @@
+use anyhow::Result;
+use chrono::DateTime;
+use csv::*;
+use num::BigUint;
+use serde::{Serialize, Serializer};
+use std::fs::{File, OpenOptions};
+use std::path::Path;
+
+#[cfg(not(test))]
+use chrono::Local;
+
+#[cfg(test)]
+use chrono::FixedOffset;
+
+#[derive(Debug, Copy, Clone, Serialize)]
+#[serde(rename_all = "UPPERCASE")]
+enum Symbol {
+    Btc,
+    Dai,
+}
+
+#[derive(Debug, Copy, Clone, Serialize)]
+enum Position {
+    Buy,
+    Sell,
+}
+
+#[derive(Debug, Clone, Serialize)]
+struct Float(String);
+
+impl From<f64> for Float {
+    fn from(float: f64) -> Self {
+        Float(float.to_string())
+    }
+}
+
+#[derive(Debug, Clone)]
+struct Integer(BigUint);
+
+impl From<BigUint> for Integer {
+    fn from(int: BigUint) -> Self {
+        Integer(int)
+    }
+}
+
+impl From<u64> for Integer {
+    fn from(int: u64) -> Self {
+        Integer(BigUint::from(int))
+    }
+}
+
+impl Serialize for Integer {
+    fn serialize<S>(&self, serializer: S) -> Result<<S as Serializer>::Ok, <S as Serializer>::Error>
+    where
+        S: Serializer,
+    {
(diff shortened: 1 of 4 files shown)


src/history.rs
```diff
@@ -0,0 +1,242 @@
...
+use num::BigUint;
+use serde::{Serialize, Serializer};
+use std::fs::{File, OpenOptions};
+use std::path::Path;
+
+#[derive(Debug, Copy, Clone, Serialize)]
+enum Symbol {
+    BTC,
+    DAI,
+}
```
luckysori, line 13: :upside_down_face: Since they're enum variants I would use PascalCase. You can then use `#[serde(rename_all = "UPPERCASE")]` to control the serialization.

src/history.rs
```diff
@@ -0,0 +1,242 @@
...
+        let expected_contents = "start_timestamp,final_timestamp,final_state,base_symbol,quote_symbol,position,base_rounded_nominal_amount,quote_rounded_nominal_amount,base_precise_amount,quote_precise_amount,peer
+2020-07-10T17:48:26.380081+10:00,2020-07-10T18:48:26.380081+10:00,Redeemed,BTC,DAI,Buy,0.01,99,1000000,99000000000000000000,QmUJF1AzhjUfDU1ifzkyuHy26SCnNHbPaVHpX1WYxYYgZg
+2020-07-10T17:48:26.380081+10:00,2020-07-10T18:48:26.380081+10:00,Redeemed,BTC,DAI,Buy,0.01,99,1000000,99000000000000000000,QmUJF1AzhjUfDU1ifzkyuHy26SCnNHbPaVHpX1WYxYYgZg
+";
+
+        assert_eq!(contents, expected_contents);
+    }
+
+    #[test]
+    fn re_use_existing_file_without_loosing_data_or_re_writing_headers() {
```
luckysori, line 215: :stuck_out_tongue: 
```suggestion
    fn re_use_existing_file_without_losing_data_or_re_writing_headers() {
```

src/history.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        Self(libp2p::PeerId::from_str("QmUJF1AzhjUfDU1ifzkyuHy26SCnNHbPaVHpX1WYxYYgZg").unwrap())
+    }
+}
+
+#[derive(Debug, Copy, Clone, Serialize)]
+enum FinalState {
+    Refunded,
+    Redeemed,
+}
```
luckysori, line 85: :question: Do we really need this? Is a trade that doesn't come to fruition worth adding to the history? Sounds like an extra feature, in case you want to keep track of fees wasted.

src/history.rs
```diff
@@ -0,0 +1,242 @@
...
+
+/// ⚠️ If you change this then you need to think about versioning
+/// All the information to write in the CVS file per trade
+#[derive(Debug, Clone, Serialize)]
+struct Trade {
```
luckysori, line 90: :question: Why do we need both `rounded` and `precise` amounts? One seems sufficient to me.

src/history.rs
D4nte: ah good point! thanks.

src/history.rs
D4nte: Let's remove.

src/history.rs
D4nte: Which one would you keep?

luckysori: That's a good question! I guess integers are more precise and for the purposes of personal accounting that would be more useful.

PR #49, opened 2020-07-13
Respawn unfinished swaps before entering main loop

luckysori:
Replacing #46 because pushing to its branch is not updating the PR lol.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,35 +2,31 @@
 #![recursion_limit = "256"]
 
 use anyhow::Context;
-use comit::{hbit, herc20};
-use futures::channel::mpsc::{Receiver, Sender};
-use futures::FutureExt;
-use futures::StreamExt;
-use futures::{Future, SinkExt};
+use futures::{
+    channel::mpsc::{Receiver, Sender},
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -264,14 +263,32 @@ fn handle_dai_balance_update(
     unimplemented!()
 }
 
+// TODO: I don't think `finished_swap` should be an Option
```
da-kami, line 266: What would `None` mean?
From the reserve point of view, we have to unreserve (and "unlock" the taker from ongoing swaps) in all these three cases (because it was reserved either way and has to be freed):

- Taker never funded (i.e. swap timed out)
- Refunded
- Redeemed

Since we have to react to all these scenarios in the same way, I wonder what `None` would represent.

luckysori:
> Is it me or we are missing the step where we save a swap in the DB?

Yeah, this PR was originally created to showcase what we would need from the database in terms of respawning. First saving to the database is a separate concern. I just wanted to get this merged so that I could continue.

src/main.rs
luckysori: AFAIK `None` means that the stream is terminated.

PR #50, opened 2020-07-13
Additional db apis

D4nte:
Finishes off the DB. There are a number of improvements/cleanup I'd like to do but I'll leave them at todo for now and focus on delivering features.


da-kami:
Is there a reason why you are not using `flush` when working with `sled`? 
I thought that each `Database` operation would end with a `flush`.

As far as I could see from the documentation `flush` this ensures that everything is actually stored on disk - and there is `flush_async` as well.

(Might be that I missed this discussion and you already know this...)

D4nte:
> Is there a reason why you are not using `flush` when working with `sled`?
> I thought that each `Database` operation would end with a `flush`.
> 
> As far as I could see from the documentation `flush` this ensures that everything is actually stored on disk - and there is `flush_async` as well.
> 
> (Might be that I missed this discussion and you already know this...)

Because `flush` is not mentioned in the readme and now I don't understand why. No example uses flush.

Indeed, I think we can revert back to making the traits async and using flush_async.

I'll do that in a follow-up PR and create a todo for it.

src/swap.rs
```diff
@@ -12,28 +12,81 @@ pub mod herc20;
...
     Herc20Hbit(SwapParams),
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct SwapParams {
     pub hbit_params: hbit::Params,
     pub herc20_params: herc20::Params,
     pub secret_hash: comit::SecretHash,
+    // TODO: Why naive and not DateTime<Local>?
```
luckysori, line 34: That's just what `btsieve` takes.

src/swap.rs
```diff
@@ -12,28 +12,81 @@ pub mod herc20;
...
+use std::sync::Arc;
 
 pub use action::{AsSwapId, BetaExpiry, BetaLedgerTime, DoItOnce, Execute, TryDoItOnce};
 pub use alice::WatchOnlyAlice;
 pub use bob::WalletBob;
 pub use db::{Database, Save};
-use std::sync::Arc;
 
-#[derive(Debug)]
+// TODO: This is awkward to manipulate
```
luckysori, line 22: You are welcome to try something different! It was just an idea :relaxed:

PR #53, opened 2020-07-15
Introduce `trade` command

D4nte:
Introduce the `trade` command. 

Also started to prepare the field to save trades in the history. I will complete that in a new PR as there is already some refactoring in there that would be good to merge to allow others to rebase now.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -1,8 +1,17 @@
 use std::path::PathBuf;
+use structopt::StructOpt;
 
-#[derive(structopt::StructOpt, Debug)]
+#[derive(StructOpt, Debug)]
 pub struct Options {
     /// Path to configuration file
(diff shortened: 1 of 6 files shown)


src/options.rs
```diff
@@ -1,8 +1,17 @@
...
     #[structopt(short = "c", long = "config", parse(from_os_str))]
     pub config_file: Option<PathBuf>,
+    /// Start trading
+    #[structopt(subcommand)]
+    pub cmd: Command,
+}
+
+#[derive(StructOpt, Debug, Copy, Clone)]
+pub enum Command {
+    Trade,
```
da-kami, line 16: ❓ Is the subcommand `Trade` or `trade`?

D4nte: It's `trade`.

PR #54, opened 2020-07-15
Implement network layer

luckysori:
Do *not* merge until we go back to depending on `comit/nectar`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2018"
 [dependencies]
 anyhow = "1"
 async-trait = "0.1"
+bimap = "0.4"
 bitcoin = { version = "0.23.0", features = ["rand", "serde"] }
 chrono = "0.4"
 clarity = "0.1"
@@ -21,7 +22,7 @@ ethereum-types = "0.9.0"
 futures = "0.3.5"
 futures-timer = "3.0.2"
 hex = "0.4"
-libp2p = { version = "0.18", default-features = false, features = ["tcp", "secio", "yamux", "mplex", "mdns", "dns"] }
+libp2p = { version = "0.21", default-features = false, features = ["tcp-tokio", "secio", "yamux", "mplex", "dns", "gossipsub", "request-response"] }
 log = "0.4"
 num = "0.3"
 num256 = "0.2"
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,28 +2,24 @@
 #![recursion_limit = "256"]
 
 use anyhow::Context;
-use chrono::{DateTime, Local, Utc};
+use chrono::{DateTime, Local};
 use futures::{
     channel::mpsc::{Receiver, Sender},
     Future, FutureExt, SinkExt, StreamExt,
 };
 use futures_timer::Delay;
 use libp2p::PeerId;
-use nectar::options::Command;
 use nectar::{
     bitcoin, config,
     config::{settings, Settings},
-    ethereum,
-    ethereum::dai,
-    history,
-    history::History,
+    ethereum::{self, dai},
+    history::{self, History},
     maker::{PublishOrders, TakeRequestDecision},
     mid_market_rate::get_btc_dai_mid_market_rate,
-    network::{self, Nectar, Orderbook, Taker},
-    options::{self, Options},
-    order::Position,
-    swap::{self, herc20, Database, SwapKind, SwapParams},
-    Maker, MidMarketRate, Spread, SwapId,
+    network::{self, Swarm, Taker},
+    options::{self, Command, Options},
+    swap::{self, Database, SwapKind},
+    Maker, MidMarketRate, Seed, Spread,
 };
 use num::BigUint;
 use std::str::FromStr;
@@ -49,7 +45,7 @@ async fn init_maker(
         .context("Could not get Dai balance")?;
 
     let btc_max_sell = maker_settings.max_sell.bitcoin;
-    let dai_max_sell = maker_settings.max_sell.dai;
+    let dai_max_sell = maker_settings.max_sell.dai.clone();
     let btc_fee_reserve = maker_settings.maximum_possible_fee.bitcoin;
 
     let initial_rate = get_btc_dai_mid_market_rate()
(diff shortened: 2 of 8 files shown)


src/bitcoin_wallet.rs
```diff
@@ -62,6 +65,20 @@ impl Wallet {
         Ok(())
     }
 
+    pub fn random_transient_sk(&self) -> anyhow::Result<SecretKey> {
```
D4nte, line 68: I guess this is for test only? If so add `#[cfg(test)]`.

If not... Shouldn't we derive it from our seed?

src/network.rs
```diff
@@ -116,16 +199,343 @@ impl Nectar {
...
+            id: order.id,
+            inner,
+            taker,
+            confirmation_channel,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct Taker(PeerId);
```
D4nte, line 429: Lol, we both did the same change but slightly different... :/

src/bitcoin_wallet.rs
luckysori: It's currently not used for testing...

> If not... Shouldn't we derive it from our seed?

In a way it is derived from the seed, since the Bitcoin wallet's private key is derived from the seed :stuck_out_tongue: Then we need to hash it with something unique so that we don't use the same one for all swaps. In `cnd` we use a swap ID, but when I call it in Nectar we still did not have one. That's why I hashed it with a random string. This function is only called once per swap anyway (then saved to a `HashMap` inside `network::Nectar`).

I think it's bad though. That's why I left the TODO just below.

Cargo.toml
```diff
@@ -7,10 +7,11 @@ edition = "2018"
 [dependencies]
 anyhow = "1"
 async-trait = "0.1"
+bimap = "0.4"
 bitcoin = { version = "0.23.0", features = ["rand", "serde"] }
 chrono = "0.4"
 clarity = "0.1"
-comit = { git = "https://github.com/comit-network/comit-rs", package = "comit", branch = "nectar" }
+comit = { git = "https://github.com/comit-network/comit-rs", package = "comit", branch = "nectar-orderbook-integration" }
```
D4nte, line 14: I assume you want to change that before merging to master, right?

src/main.rs
```diff
@@ -196,49 +185,68 @@ async fn execute_swap(
...
     ethereum_wallet: Arc<ethereum::Wallet>,
     bitcoin_connector: Arc<comit::btsieve::bitcoin::BitcoindConnector>,
     ethereum_connector: Arc<comit::btsieve::ethereum::Web3Connector>,
-    sender: Sender<FinishedSwap>,
+    finished_swap_sender: Sender<FinishedSwap>,
 ) {
     match network_event {
-        network::Event::TakeRequest(order) => {
+        network::Event::TakeOrderRequest(order) => {
             // decide & take & reserve
```
D4nte, line 198: Is this comment still valid?

src/main.rs
```diff
@@ -196,49 +185,68 @@ async fn execute_swap(
...
+                            };
                         }
                         Err(e) => tracing::error!("Error when trying to create new order: {}", e),
                     }
                 }
                 Ok(TakeRequestDecision::RateNotProfitable)
                 | Ok(TakeRequestDecision::InsufficientFunds)
                 | Ok(TakeRequestDecision::CannotTradeWithTaker) => {
-                    swarm.orderbook.ignore(order);
+                    swarm.deny(order);
```
D4nte, line 220: 👍

src/network.rs
luckysori: Stuck with mine after merging because it was easier, but we can change it back to yours.

src/main.rs
```diff
@@ -257,8 +265,13 @@ fn handle_rate_update(
                     new_sell_order,
                     new_buy_order,
                 })) => {
-                    swarm.orderbook.publish(new_sell_order.into());
-                    swarm.orderbook.publish(new_buy_order.into());
+                    if let Err(e) = swarm.make(new_sell_order.into()) {
+                        tracing::error!("Failed to make new order: {}", e)
```
D4nte, line 269: I think the terminology is confusing here. We did not fail to make (generate) an order, we failed to publish it.

src/main.rs
```diff
@@ -277,12 +290,14 @@ fn handle_rate_update(
...
-    swarm: &mut libp2p::Swarm<Nectar>,
+    swarm: &mut Swarm,
 ) {
     match btc_balance_update {
         Ok(btc_balance) => match maker.update_bitcoin_balance(btc_balance) {
             Ok(Some(new_sell_order)) => {
-                swarm.orderbook.publish(new_sell_order.into());
+                if let Err(e) = swarm.make(new_sell_order.into()) {
+                    tracing::error!("Failed to make new order: {}", e)
+                };
```
D4nte, line 300:
```suggestion
                let _ = swarm.make(new_sell_order.into()).map_err(|e| {
                    tracing::error!("Failed to make new order: {}", e)
                });
```

Do you really prefer the `if let Err(e) = function() { log(e) }` syntax? I find it very verbose as a reader. It focused the attention to the matching part `Err(e)` which is just about logging an error, taking it away from the actual function call `swarm.make` which is the important part of the line.

src/bitcoin_wallet.rs
D4nte: Yeah, thinking about it, I think we should just use bip32 to derive those keys. Maybe under a different account or path than what bitcoind does.

src/network.rs
```diff
@@ -1,76 +1,482 @@
...
+
+        self.inner.announce_trading_pair(TradingPair {
+            buy: SwapType::Herc20,
+            sell: SwapType::Hbit,
+        })?;
+
+        Ok(())
+    }
+
+    pub fn make(&mut self, order: PublishOrder) -> anyhow::Result<()> {
```
D4nte, line 96:
```suggestion
    pub fn publish(&mut self, order: PublishOrder) -> anyhow::Result<()> {
```

Cargo.toml
luckysori: Yeah, I was about to block this PR on this.

src/main.rs
luckysori: Haha, let's remove it.

src/main.rs
luckysori: Fair, I'll change that.

src/network.rs
D4nte: Mine was just to make it easy for you to add the fields that you need :)

src/main.rs
luckysori: Oh, yeah, I forgot that we could do it with `map_err`. That's what I was looking for.

src/bitcoin_wallet.rs
luckysori: That sounds very good! I would say we should do that in a separate PR though.

src/network.rs
```diff
@@ -83,64 +489,100 @@ impl From<BtcDaiOrder> for PublishOrder {
...
 }
 
-#[cfg(test)]
-impl Default for Order {
-    fn default() -> Self {
-        Self {
-            inner: BtcDaiOrder::default(),
-            taker: Taker::default(),
+// TODO: The comit::network::orderbook expects a token contract but we
+// assume DAI. Maybe we should use a type that combines dai::Amount
```
D4nte, line 493: We should do whatever solution is best for us to handle several dai networks and their contract address. Please do not consider other ERC20 tokens when designing this.

src/swap.rs
```diff
@@ -25,6 +25,14 @@ pub enum SwapKind {
     Herc20Hbit(SwapParams),
 }
 
+impl From<SwapKind> for SwapParams {
```
D4nte, line 28: Because you lose information by doing this conversion I'd been more keen to see
```
impl SwapKind {
	fn params(self) -> SwapParams {
...
	}
}
```

It's a nitpick.

src/network.rs
luckysori: Good point.

src/swap.rs
luckysori: Fair enough!

src/network.rs
luckysori: Reverted it back to your design :D

src/swap.rs
luckysori: Never used anyway. I'm just gonna remove it.

PR #58, opened 2020-07-17
Generate the Bitcoin root extended private key from the seed

D4nte:
This is done the same manner than bitcoind does. Which means
that the seed can later be imported directly in bitcoind for
recovery purposes.

A test is done to prove it.

### DIFF
--- a/src/bitcoin/wallet.rs
+++ b/src/bitcoin/wallet.rs
@@ -215,6 +215,7 @@ impl Wallet {
         Ok(())
     }
 
+    // TODO: Just hash the seed instead of the public key of the seed (as a private key)
     fn gen_name(private_key: PrivateKey) -> String {
         let mut hash_engine = PubkeyHash::engine();
         private_key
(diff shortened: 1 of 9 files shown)


src/bitcoin/wallet.rs
```diff
@@ -71,7 +77,7 @@ impl Wallet {
         let mut random_bytes = [0u8; 32];
 
         rand::thread_rng().fill_bytes(&mut random_bytes);
-
+        // TODO: use bitcoin_hashes instead of adding new dependency sha2
```
luckysori, line 80: :+1:

src/bitcoin/wallet.rs
```diff
@@ -1,27 +1,33 @@
...
+    seed: Seed,
     network: Network,
 }
 
 impl Wallet {
     pub async fn new(seed: Seed, url: Url, network: Network) -> anyhow::Result<Wallet> {
-        let key = seed.secret_key()?;
+        let key = secp256k1::SecretKey::from_slice(&seed.bytes())?;
 
         let private_key = ::bitcoin::PrivateKey {
```
luckysori, line 32: :wrench: Now that the wallet just holds the seed, it's really weird to generate a private key just for the purpose of generating a name for the wallet.

PR #59, opened 2020-07-17
Taker persistence

luckysori:
Patch https://github.com/coblox/nectar/commit/9a96fe45b9135a21e7a60cb5981bccd5275df078 fixes a bug where we were calling the opposite function in the one implemented `SwapKind` branch when executing swaps (Thanks, @tcharding).

The rest of the work (in https://github.com/coblox/nectar/commit/75235bf60231f8951df835e9dbfc1d5618596699) was done to ensure that the logic of keeping a maximum of 1 active order/swap per taker is also applied to swaps that have been respawned after restarting Nectar.

To achieve that we save the `ActiveTakers` (renamed from `PeersWithOngoingSwaps`) to a database and we add the taker's `PeerId` to `SwapParams`.

The `ActiveTakers` functionality has moved to the network layer. This also allows us to more readily ignore `TakeOrderRequests` from active takers.


src/network.rs
```diff
@@ -177,13 +194,21 @@ impl Nectar {
...
+        self.active_takers.insert(order.taker)?;
+
         self.orderbook.confirm(order.id, order.confirmation_channel);
+
+        Ok(())
     }
 
-    fn deny(&mut self, order: TakenOrder) {
+    fn deny(&mut self, order: TakenOrder) -> anyhow::Result<()> {
+        self.active_takers.remove(&order.taker)?;
```
da-kami, line 206: ❓ Is this still needed. With the logic of `active_takers` being moved to the network layer  there should not be an error that leads to the a taker being stored but then running into deny. (It does not harm us to do it at this point, thus I only raised a question.)

src/network.rs
```diff
@@ -240,21 +269,44 @@ impl NetworkBehaviourEventProcess<orderbook::BehaviourOutEvent> for Nectar {
                 response_channel,
                 order_id,
             } => {
+                let taker = Taker::new(taker_peer_id);
+                let ongoing_trade_with_taker_exists = match self.active_takers.contains(&taker) {
+                    Ok(res) => res,
+                    Err(e) => {
+                        tracing::warn!(
```
D4nte, line 276: This is more appropriate as an error because you cannot process more orders due to a DB issue.

luckysori: Good point, thanks for reviewing the details :relaxed:

luckysori, line 276:
```suggestion
                        tracing::error!(
```

src/network.rs
```diff
@@ -240,21 +269,44 @@ impl NetworkBehaviourEventProcess<orderbook::BehaviourOutEvent> for Nectar {
                 response_channel,
                 order_id,
             } => {
+                let taker = Taker::new(taker_peer_id);
+                let ongoing_trade_with_taker_exists = match self.active_takers.contains(&taker) {
```
da-kami, line 273: Neat! So, we only bubble the event up to handle it in the maker if there is no active trade with this taker.

luckysori: Yes! It took me quite long to reach this design, but I think it makes sense to move this logic here. And we get the benefit that you describe :)

src/network.rs
luckysori: You're right, we don't need this :+1:

luckysori: I'll fix it with PR #62.

luckysori: Nice catch!

PR #60, opened 2020-07-17
Implement `nectar walletinfo` command

D4nte:


src/bitcoin/wallet.rs
```diff
@@ -161,6 +161,36 @@ impl Wallet {
...
+                .get_descriptor_info(&descriptor)
+                .await?;
+            let descriptor = format!("{}#{}", descriptor, response.checksum);
+            descriptors.push(descriptor);
+        }
+
+        Ok(descriptors)
+    }
+
+    fn hd_paths(&self) -> Vec<&str> {
```
luckysori, line 190: ❓ How come there are two? Are these both used by bitcoind? Could/should we define a constant instead? It would help to clarify with a docstring.

src/bitcoin/bitcoind.rs
```diff
@@ -230,6 +230,37 @@ impl Client {
         Ok(())
     }
 
+    pub async fn derive_addresses(
```
luckysori, line 233: :question: Is this one just meant to be used for testing?

D4nte: We only use it in test BUT it is a valid command for testnet and mainnet bitcoind. This is why I don't put it behind a `#[cfg(test)]` compiler attribute.

The reason why I put the `generate_to_address` function behind `#[cfg(test)]` is because this command is only valid on regtest.

PR #61, opened 2020-07-17
Clean up lib.rs

D4nte:
To later remove it.


src/history.rs
```diff
@@ -9,7 +9,7 @@ use std::path::Path;
 #[cfg(not(test))]
 use chrono::Local;
 
-use crate::ensure_directory_exists;
+use crate::fs::ensure_directory_exists;
```
luckysori, line 12: 🙃 Move it with the rest of the use statements.

PR #62, opened 2020-07-17
Reserve funds before respawn

luckysori:
Given that we only free funds once the swap execution is over, we need to make sure that we always reserve funds before respawning (since the `Maker`'s state is lost between restarts of Nectar).

Additionally, we could examine swaps and decide whether we have enough money or not to continue with the execution successfully, but it would be quite complex. Instead, we let the swap execution fail, which is functionally equivalent to preventing the respawn.

---

There is a second patch which makes it so that `deny` doesn't try to remove the order's taker from the set of `ActiveTaker`s, as @da-kami  correctly suggested in his review of PR #59.

### DIFF
--- a/src/command/trade.rs
+++ b/src/command/trade.rs
@@ -1,4 +1,4 @@
-#![allow(unreachable_code, unused_variables, clippy::unit_arg)]
+#![allow(unreachable_code, unused_variables)]
 
 use crate::{
     bitcoin,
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -599,19 +592,33 @@ async fn trade(
...
-        // the state of the swap here.
+        // Reserve funds
+        match swap {
+            SwapKind::HbitHerc20(SwapParams {
+                swap_id,
+                ref herc20_params,
+                ..
+            }) => {
+                let fund_amount = herc20_params.asset.clone().into();
+                maker.dai_reserved_funds = maker.dai_reserved_funds.clone() + fund_amount;
```
da-kami, line 611: Note: Added follow-up todo to contain this logic in once place in the maker.

PR #66, opened 2020-07-20
Initialise connectors

da-kami:


da-kami:
introduced this on top of the re-structuring again.

PR #67, opened 2020-07-20
Move trade to separate module

da-kami:


src/history.rs
```diff
@@ -130,7 +130,7 @@ impl Serialize for LocalDateTime {
 }
 
 /// All the information to write in the CVS file per trade
-// ⚠️ If you change this then you need to think about versioning
+// If you change this then you need to think about versioning
```
da-kami, line 133: ⚠️ This caused emacs problems

src/command/trade.rs
```diff
@@ -0,0 +1,593 @@
...
+            dai_balance_update = dai_balance_update_receiver.next().fuse() => {
+                handle_dai_balance_update(dai_balance_update.unwrap(), &mut maker, &mut swarm);
+            }
+        }
+    }
+
+    Ok(())
+}
+
+const ENSURED_CONSUME_ZERO_BUFFER: usize = 0;
```
D4nte, line 132: Does this need to be a const? Top of file might be more appropriate

luckysori:
> Looks good. @luckysori are you happy with the compiler attribute move?

I don't mind :)

PR #68, opened 2020-07-20
Ask node for gas limit estimate for Wallet::send_transaction

luckysori:
- Introduce ethereum::Wallet::gas_limit to do so.
- Remove unused geth::send_transaction. We always sign the transaction
  locally and broadcast a raw transaction.

### DIFF
--- a/src/ethereum/geth.rs
+++ b/src/ethereum/geth.rs
@@ -8,7 +8,8 @@ use comit::{
 };
 use ethereum_types::U256;
 use num::{BigUint, Num};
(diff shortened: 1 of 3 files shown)


src/ethereum/geth.rs
```diff
@@ -40,20 +41,6 @@ impl Client {
         Ok(chain_id)
     }
 
-    pub async fn send_transaction(&self, request: SendTransactionRequest) -> anyhow::Result<Hash> {
```
D4nte, line 43: 📣 This is needed to withdraw ether!

PR #69, opened 2020-07-20
Initialise connectors and `trade` test

da-kami:
See latest commit message for more info :)


src/network.rs
```diff
@@ -566,8 +566,9 @@ impl From<PublishOrder> for comit::network::orderbook::NewOrder {
             } => Self {
                 buy: base.into(),
                 sell: asset::Erc20::new(
-                    // TODO: Handle other networks
-                    DaiContractAddress::from_public_chain_id(ChainId::regtest())
+                    // FIXME: The dai amount does neither know about the contract address not the network atm
```
D4nte, line 569: This TODO seems to be critical, can we do it now?

src/maker.rs
```diff
@@ -298,7 +304,7 @@ mod tests {
         bitcoin::Amount::from_btc(btc).unwrap()
     }
 
-    fn dai(dai: f64) -> dai::Amount {
+    fn dai_amount(dai: f64) -> dai::Amount {
```
da-kami, line 307: 🙃  I'd change this to `dai(...) -> dai::Asset`, the point of this function was that we don't have to care what is behind it.

PR #70, opened 2020-07-20
Preparing the field for withdraw command

D4nte:
The patches can be reviewed individually.

### DIFF
--- a/src/bitcoin/wallet.rs
+++ b/src/bitcoin/wallet.rs
@@ -58,20 +58,31 @@ impl Wallet {
 
         // We assume the wallet present with the same name has the
         // same seed, which is fair but could be safer.
-        if info.is_err() {
-            // TODO: Probably need to protect the wallet with a passphrase
-            self.bitcoind_client
-                .create_wallet(&self.name, None, Some(true), None, None)
-                .await?;
+        match info {
+            Err(_) => {
+                // TODO: Probably need to protect the wallet with a passphrase
+                self.bitcoind_client
+                    .create_wallet(&self.name, None, Some(true), None, None)
+                    .await?;
 
-            let wif = self.seed_as_wif();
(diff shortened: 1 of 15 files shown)


src/bitcoin/wallet.rs
```diff
@@ -58,20 +58,31 @@ impl Wallet {
...
+                let wif = self.seed_as_wif();
+
+                self.bitcoind_client
+                    .set_hd_seed(&self.name, Some(true), Some(wif))
+                    .await
+            }
+            Ok(WalletInfoResponse {
+                hd_seed_id: None, ..
+            }) => {
+                // The wallet may have been previously created the `sethdseed` call may have failed
```
luckysori, line 77: ❓ 
```suggestion
                // The wallet may have been previously created, but the `sethdseed` call may have failed
```

src/config/serde/ethereum_address.rs
```diff
@@ -0,0 +1,67 @@
+use crate::swap::ethereum;
+use serde::{de, export::fmt, Deserialize, Deserializer, Serializer};
+use std::str::FromStr;
+
+pub fn deserialize<'de, D>(deserializer: D) -> Result<Option<ethereum::Address>, D::Error>
```
da-kami, line 5: ❓ Why do we need a custom deserialiser for this now? Isn't the one from the `comit` library sufficient?

D4nte: It is surprisingly not. Just take my test change and try it locally ;)
Also, our test to use `toml::from_str` instead of using `config_rs` like it is done in the `main` was useless.

src/bitcoin/wallet.rs
D4nte: yes thanks

PR #71, opened 2020-07-21
Implement `nectar withdraw` for Ether and Bitcoin

D4nte:
Almost there, getting the following error:
```
thread 'command::withdraw::tests::withdraw_command' panicked at 'called `Result::unwrap()` on an `Err` value: failed to get gas price

Caused by:
    0: JSON-RPC request Request { id: "1", jsonrpc: "2.0", method: "eth_estimateGas", params: [Object({"gas_price": String("1"), "to": String("0x0000000000000000000000000000000000000000"), "value": String("0x214e8348c4f00000")})] } failed
    1: JSON-RPC request failed with code -32000: gas required exceeds allowance (6320079) or always failing transaction', src/command/withdraw.rs:109:22
```

### DIFF
--- a/src/bitcoin.rs
+++ b/src/bitcoin.rs
@@ -5,7 +5,7 @@ mod bitcoind;
 mod wallet;
 
 use crate::float_maths::string_int_to_float;
-pub use ::bitcoin::{Address, Network};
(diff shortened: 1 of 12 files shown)


D4nte:
Will re-open from a new branch

PR #76, opened 2020-07-22
Invert the rate

D4nte:
The rate is always the same, no matter the position so when we want
to convert from quote to base (dai to btc) we needed to divide by the
rate, not multiply.

### DIFF
--- a/src/ethereum/dai.rs
+++ b/src/ethereum/dai.rs
@@ -10,7 +10,7 @@ use comit::{
 };
 use conquer_once::Lazy;
 use ethereum_types::U256;
-use num::{pow::Pow, BigUint, CheckedAdd, ToPrimitive, Zero};
+use num::{pow::Pow, BigUint, CheckedAdd, Integer, ToPrimitive, Zero};
 use std::str::FromStr;
 
 pub const ATTOS_IN_DAI_EXP: u16 = 18;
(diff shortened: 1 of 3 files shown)


src/ethereum/dai.rs
```diff
@@ -267,36 +271,36 @@ mod tests {
...
     #[test]
     fn worth_in_result_truncated_1() {
-        let dai = Amount::from_dai_trunc(101.0).unwrap();
-        let rate = Rate::try_from(0.000_123_456).unwrap();
+        let dai = Amount::from_dai_trunc(112.648125).unwrap();
+        let rate = Rate::try_from(9125.0).unwrap();
 
         let res: bitcoin::Amount = dai.worth_in(rate).unwrap();
 
         // Result is 0.012469056 btc or 1246905.6 satoshis
```
da-kami, line 290:
```suggestion
```

or update the values.

src/ethereum/dai.rs
```diff
@@ -267,36 +271,36 @@ mod tests {
...
     #[test]
     fn worth_in_result_truncated_2() {
-        let dai = Amount::from_dai_trunc(100_001.0).unwrap();
-        let rate = Rate::try_from(0.000_001_234).unwrap();
+        let dai = Amount::from_dai_trunc(0.01107).unwrap();
+        let rate = Rate::try_from(9000.0).unwrap();
 
         let res: bitcoin::Amount = dai.worth_in(rate).unwrap();
 
         // Result is 12,340,123.4 satoshis
```
da-kami, line 302:
```suggestion
```

src/maker.rs
```diff
@@ -730,4 +730,19 @@ mod tests {
         assert_eq!(result, TakeRequestDecision::GoForSwap);
         assert_eq!(maker.dai_reserved_funds, dai_amount(1.0))
     }
+
+    #[test]
+    fn new_buy_order_is_correct() {
```
da-kami, line 735: 🙃 Shall we add one for the sell order as well?

PR #77, opened 2020-07-22
Print newly created orders

da-kami:
Minimal printing of the order position and nominal values.

I think this is actually so basic that it is fine to add. We can also drop it, but I feel a bit more context would be good when running `trade`. (I also thought about printing the `Rate` and `Balance` but did not go there yet...)

```
Jul 22 15:50:15.919  INFO nectar::trace: Initialized tracing with level: INFO
Jul 22 15:50:15.923  INFO nectar::config::seed: Read in seed from file: /Users/dakami/Library/Application Support/nectar/seed.pem
Jul 22 15:50:16.471  INFO libp2p_gossipsub::behaviour: Subscribed to topic: BTC/DAI    
Jul 22 15:50:16.472  INFO nectar::maker: New order created: Position: Sell, BTC: 0.09990725, DAI: 967.83
Jul 22 15:50:16.472  INFO nectar::maker: New order created: Position: Buy, BTC: 175294, DAI: 20
Jul 22 15:50:16.473  INFO libp2p_gossipsub::behaviour: Published message: MessageId("2276897037252766832")    
Jul 22 15:50:16.473  INFO libp2p_gossipsub::behaviour: Published message: MessageId("13304717341951990730")  
```

Note that the `Buy` order's BTC amount is already fixed, here https://github.com/coblox/nectar/pull/76#pullrequestreview-453010128

### DIFF
--- a/src/order.rs
+++ b/src/order.rs
@@ -6,7 +6,8 @@ use crate::{
     },
     MidMarketRate, Rate, Spread,
 };
-use std::{cmp::min, convert::TryFrom};
+use std::fmt::{Display, Formatter};
+use std::{cmp::min, convert::TryFrom, fmt};
 
 #[derive(Debug, Copy, Clone, strum_macros::Display)]
 #[strum(serialize_all = "UPPERCASE")]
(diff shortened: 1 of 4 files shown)


src/order.rs
```diff
@@ -145,6 +146,19 @@ impl BtcDaiOrder {
     }
 }
 
+impl Display for BtcDaiOrder {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "Position: {}, BTC: {}, DAI: {}",
+            self.position,
+            self.base.amount.as_btc(),
```
D4nte, line 155: This is not needed, the amount already implement a proper `std::fmt::Display` that includes the currency symbol.

da-kami:
> I could be totally wrong on this, but it seems weird to me that we would log the creation of the order, rather than its publication. After all, one could call this function and never actually do anything useful with it (publish it), which could cause confusion.

Fair point, I moved the printing to the network code when the order is published.

da-kami:
Finally! 😅

PR #79, opened 2020-07-22
Maker integration test

da-kami:
I am not sure if such tests add more much more than showcasing the maker. I feel we are not testing much more with this test, and it will have to be adapted once the logic changes (together with the more isolated logic tests).

I see more value in writing (separate) e2e tests (from creating an order till finishing swap execution), but for that we need a simulated taker which is not trivial.

@D4nte please let me know what you think.


da-kami:
sorry, was meant to be Draft PR

D4nte:
You know you can make open PR draft PRs?

PR #80, opened 2020-07-22
Maker integration test

da-kami:
I am not sure if such tests add more much more than showcasing the maker. I feel we are not testing much more with this test, and it will have to be adapted once the logic changes (together with the more isolated logic tests).

I see more value in writing (separate) e2e tests (from creating an order till finishing swap execution), but for that we need a simulated taker which is not trivial.

@D4nte please let me know what you think.

### DIFF
--- a/src/maker.rs
+++ b/src/maker.rs
@@ -746,3 +746,127 @@ mod tests {
         assert_eq!(new_buy_order.quote.amount, dai_amount(18.0));
     }
 }
+
+#[cfg(test)]
+mod integration_tests {
+    use super::*;
+    use crate::ethereum::ChainId;
+    use crate::rate::Rate;
+    use libp2p::PeerId;
+    use std::convert::TryFrom;
+
+    fn btc(btc: f64) -> bitcoin::Amount {
+        bitcoin::Amount::from_btc(btc).unwrap()
+    }
+
+    fn btc_asset(amount: f64) -> bitcoin::Asset {
+        bitcoin::Asset {
+            amount: btc(amount),
+            network: bitcoin::Network::Bitcoin,
+        }
+    }
+
+    fn dai_amount(dai: f64) -> dai::Amount {
+        dai::Amount::from_dai_trunc(dai).unwrap()
+    }
+
+    fn some_btc(btc: f64) -> Option<bitcoin::Amount> {
+        Some(bitcoin::Amount::from_btc(btc).unwrap())
+    }
+
+    fn some_dai(dai: f64) -> Option<dai::Amount> {
+        Some(dai::Amount::from_dai_trunc(dai).unwrap())
+    }
+
+    fn some_rate(rate: f64) -> Option<MidMarketRate> {
+        Some(MidMarketRate::new(Rate::try_from(rate).unwrap()))
+    }
+
+    fn spread(spread: u16) -> Spread {
+        Spread::new(spread).unwrap()
+    }
+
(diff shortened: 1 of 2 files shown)


src/maker.rs
```diff
@@ -746,3 +746,127 @@ mod tests {
...
+            amount: btc(amount),
+            network: bitcoin::Network::Bitcoin,
+        }
+    }
+
+    fn dai_amount(dai: f64) -> dai::Amount {
+        dai::Amount::from_dai_trunc(dai).unwrap()
+    }
+
+    fn some_btc(btc: f64) -> Option<bitcoin::Amount> {
```
D4nte, line 773: Instead of re-defining these functions, move them to the test harness module.

src/maker.rs
```diff
@@ -746,3 +746,127 @@ mod tests {
...
+            position: Position::Sell,
+            base: btc_asset(1.0),                  // deduct fee
+            quote: dai_asset(dai_amount(9_450.0)), // profitable_rate = 9450, quote_amount = 0.9999 * 9450
+        };
+
+        assert_eq!(initial_buy_order, expected_initial_buy_order);
+        assert_eq!(initial_sell_order, expected_initial_sell_order);
+
+        // take request for the initial SELL order
+        let taker_order = TakenOrder {
```
luckysori, line 833: 🔧 Just rename this variable to `taken_sell_order` or even `taken_initial_sell_order` and you don't need the comment.

src/maker.rs
```diff
@@ -746,3 +746,127 @@ mod tests {
         assert_eq!(new_buy_order.quote.amount, dai_amount(18.0));
     }
 }
+
+#[cfg(test)]
+mod integration_tests {
```
luckysori, line 751: :upside_down_face: This doesn't look like an integration test to me :stuck_out_tongue:

src/maker.rs
```diff
@@ -746,3 +746,127 @@ mod tests {
...
+        };
+
+        let result = maker.process_taken_order(taker_order).unwrap();
+
+        assert_eq!(result, TakeRequestDecision::GoForSwap);
+
+        let new_sell_order = maker.new_sell_order().unwrap();
+        let new_buy_order = maker.new_buy_order().unwrap();
+
+        // We still have enough more funds than max available on both sides, thus the orders should still be the same.
```
luckysori, line 845:
```suggestion
        // We still have more funds than max available on both sides, thus the orders should still be the same.
```

src/maker.rs
luckysori: In general, I don't see the difference between this test and the ones above, other than the number of assertions.

src/maker.rs
```diff
@@ -746,3 +746,127 @@ mod tests {
...
```
luckysori, line 862:
```suggestion
        // We still have more funds than max available on both sides, thus the orders should still be the same.
```

src/maker.rs
da-kami: Yeah, that's why I stopped writing the test.
I don't think the test adds more value than the more fine-granular tests above, besides the fact that it showcases Nectar.

da-kami:
After giving this some thought I would not focus on writing this test now.
Besides showcasing the application the test does not add much value. In case we actually do test side-effects of functions being called in a certain order, it would be hard to define what does side-effect are concretely. 

I would not invest more time into this test. We can, however, re-evaluate if more fine-granular tests for certain functions are needed.

Please ping me if you disagree @D4nte @luckysori

PR #82, opened 2020-07-23
Generate Ethereum Private key using BIP32/BIP44

D4nte:


src/ethereum/wallet.rs
```diff
@@ -16,6 +16,16 @@ use num::BigUint;
 use num256::Uint256;
 use url::Url;
 
+use anyhow::anyhow;
```
luckysori, line 19: :upside_down_face: I prefer it if all the `use` statements are adjacent and the `pub use` statements appear in a separate block.

src/ethereum/wallet.rs
```diff
@@ -137,6 +165,7 @@ impl Wallet {
         chain_id: ChainId,
     ) -> anyhow::Result<Hash> {
         self.assert_chain(chain_id).await?;
+        let private_key = self.private_key();
```
luckysori, line 168: :question: I don't see why this needed changing.

src/ethereum/wallet.rs
```diff
@@ -70,6 +72,32 @@ impl Wallet {
         }
     }
 
+    pub fn private_key_from_seed(seed: &Seed) -> anyhow::Result<clarity::PrivateKey> {
+        let private_key = Self::root_extended_private_key_from_seed(seed)?
```
luckysori, line 76: :question: How do we know what `N` we are using when calling `root_extended_private_key_from_seed`? I am surprised that this compiles :open_mouth:

src/ethereum/wallet.rs
D4nte: me too :)

src/ethereum/wallet.rs
D4nte: :) The `N` does not matter to me so I am happy it compiles. I thought the same than you.

src/ethereum/wallet.rs
D4nte: It's a function call now.

D4nte: But actually it's an unnecessary change, now fixed.

PR #85, opened 2020-07-27
Remove lib.rs and some linting fix

D4nte:


src/swap/bitcoin.rs
```diff
@@ -101,13 +101,7 @@ impl hbit::ExecuteRefund for Wallet {
 }
 
 impl Wallet {
-    pub async fn redeem(
```
luckysori, line 104: Interesting. This has uncovered a little problem with implementations of `ExecuteRefund`. Thank you! I'll open a PR for that soon.

PR #87, opened 2020-07-28
Wait until transaction receipt is fetched instead of sleeping

luckysori:


src/ethereum/wallet.rs
```diff
@@ -290,28 +308,43 @@ impl Wallet {
...
-    pub async fn get_transaction_receipt(
+    pub async fn wait_until_transaction_receipt(
         &self,
         transaction_hash: Hash,
     ) -> anyhow::Result<TransactionReceipt> {
-        self.geth_client
-            .get_transaction_receipt(transaction_hash)
-            .await
+        let start_time = std::time::Instant::now();
+        let max_retry_time = Duration::from_millis(60_000);
```
da-kami, line 330: ❓ Is maximum retry time arbitrarily chosen?

luckysori: Yep. I thought about making it configurable, but at this stage it seems excessive. Maybe we don't need a `max_retry_time` at all though. I just want to prevent this waiting forever, but maybe that's also over-engineered.

PR #89, opened 2020-07-29
Tackle some TODOs in Bitcoin wallet

luckysori:


src/network.rs
```diff
@@ -123,10 +123,37 @@ pub enum Event {
...
 }
 
+impl SwapMetadata {
+    pub fn swap_id(&self) -> SwapId {
+        self.swap_id
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct SwapId {
```
da-kami, line 137: 🙃 As far as I can see we are still just using the `SharedSwapId`, not sure if we have to care about the "inner" one.
Or did I miss something?

src/bitcoin/wallet.rs
```diff
@@ -84,18 +72,16 @@ impl Wallet {
         }
     }
 
-    pub fn random_transient_sk(&self) -> anyhow::Result<SecretKey> {
-        // TODO: Replace random bytes with SwapId or SharedSwapId?
-        let mut random_bytes = [0u8; 32];
+    pub fn derive_transient_sk(&self, swap_id: SwapId) -> SecretKey {
```
D4nte, line 75: I think it'd be simpler and less "make your own" crypto if we were just deriving new keys from the seed under a different path. Let's say '2'.

luckysori: Oh, yeah. I didn't even think we could do that. That's a better approach.

src/network.rs
luckysori: We use the inner one to derive Bitcoin transient keys from it.

luckysori:
Moving forward with this. I created an internal post to discuss @D4nte's alternative solution :smiley:

PR #91, opened 2020-07-29
Fix most TODOs I wrote

luckysori:
Each patch should be self-explanatory. I recommend to review them one by one.

### DIFF
--- a/src/command/withdraw.rs
+++ b/src/command/withdraw.rs
@@ -19,7 +19,7 @@ pub async fn withdraw(
         }
         Withdraw::Dai { amount, to_address } => {
             let tx_id = ethereum_wallet
-                .transfer_dai(to_address, amount.clone(), ethereum_wallet.chain_id)
+                .transfer_dai(to_address, amount.clone(), ethereum_wallet.chain_id())
                 .await?;
             Ok(format!(
                 "{} transferred to {}\nTransaction id: {}",
@@ -33,7 +33,7 @@ pub async fn withdraw(
                     amount.clone(),
                     Some(STANDARD_ETH_TRANSFER_GAS_LIMIT),
                     None,
-                    ethereum_wallet.chain_id,
+                    ethereum_wallet.chain_id(),
                 )
                 .await?;
             Ok(format!(
@@ -83,8 +83,7 @@ mod tests {
         let ethereum_wallet = crate::ethereum::Wallet::new(
             seed,
             ethereum_blockchain.node_url.clone(),
-            ethereum_blockchain.token_contract().unwrap(),
-            ChainId::regtest(),
+            ethereum::Chain::new(ChainId::regtest(), ethereum_blockchain.token_contract()),
         )
         .await
         .unwrap();
@@ -103,7 +102,7 @@ mod tests {
                 ethereum_address,
                 Erc20 {
                     quantity: Erc20Quantity::from_wei(10_000_000_000_000_000_000u64),
-                    token_contract: ethereum_blockchain.token_contract().unwrap(),
+                    token_contract: ethereum_blockchain.token_contract(),
                 },
                 ChainId::regtest(),
             )
--- a/src/ethereum/dai.rs
+++ b/src/ethereum/dai.rs
@@ -1,12 +1,12 @@
-use crate::float_maths::string_int_to_float;
 use crate::{
     bitcoin::{self, SATS_IN_BITCOIN_EXP},
-    float_maths::{divide_pow_ten_trunc, multiply_pow_ten, truncate},
+    ethereum,
+    float_maths::{divide_pow_ten_trunc, multiply_pow_ten, string_int_to_float, truncate},
     Rate,
 };
 use comit::{
     asset::{ethereum::FromWei, Erc20, Erc20Quantity},
-    ethereum::{Address, ChainId},
+    ethereum::Address,
 };
 use conquer_once::Lazy;
 use ethereum_types::U256;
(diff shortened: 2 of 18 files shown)


src/mid_market_rate.rs
```diff
@@ -64,8 +64,10 @@ mod kraken {
         type Error = anyhow::Error;
 
         fn try_from(AskAndBid { ask, bid }: AskAndBid) -> anyhow::Result<Self> {
-            // TODO: Fix the ceil to precision of 9
-            let value = ((bid + ask) / 2f64).ceil();
+            let value = (bid + ask) / 2f64;
```
D4nte, line 67: Should we try to avoid this floating arithmetics?

src/ethereum/dai.rs
```diff
@@ -46,47 +46,67 @@ static DAI_CONTRACT_ADDRESS_ROPSTEN: Lazy<Address> = Lazy::new(|| {
...
-impl From<DaiContractAddress> for Address {
-    fn from(var: DaiContractAddress) -> Self {
-        match var {
-            DaiContractAddress::Mainnet => *DAI_CONTRACT_ADDRESS_MAINNET,
-            DaiContractAddress::Kovan => *DAI_CONTRACT_ADDRESS_KOVAN,
-            DaiContractAddress::Rinkeby => *DAI_CONTRACT_ADDRESS_RINKEBY,
-            DaiContractAddress::Ropsten => *DAI_CONTRACT_ADDRESS_ROPSTEN,
-            DaiContractAddress::Local(address) => address,
+impl Chain {
+    pub fn new(chain_id: ChainId, dai_contract_address: Address) -> Self {
```
D4nte, line 65: That's an interesting design. I think I like it.

src/ethereum/wallet.rs
```diff
@@ -31,45 +29,41 @@ const DERIVATION_PATH: EthereumDerivationPath<wagyu_ethereum::network::Mainnet>
...
 }
 
 impl Wallet {
-    pub async fn new(
-        seed: Seed,
-        url: Url,
-        dai_contract_addr: Address,
-        chain_id: ChainId,
-    ) -> anyhow::Result<Self> {
+    pub async fn new(seed: Seed, url: Url, chain: ethereum::Chain) -> anyhow::Result<Self> {
```
D4nte, line 36: This `ethereum::Chain` is specific to `dai` (and defined in `dai`).

I think we are getting the module boundaries wrong here.

If this chain is for ethereum and it should be defined in the `ethereum` module.

I know it has `DAI` characteristics but I think it's fine has `dai` is a submodule of `ethereum`. `ethereum` can bring together `dai` and `ethereum` concepts.

I would try to move the `Chain` definition and `impl` block in `ethereum` and have some logic around the address defined in `dai` and brought in `ethereum`.

Not sure it makes sense, if it does not, let me know and I can give it a whack.

src/ethereum/wallet.rs
```diff
@@ -31,45 +29,41 @@ const DERIVATION_PATH: EthereumDerivationPath<wagyu_ethereum::network::Mainnet>
...
         private_key: clarity::PrivateKey,
         url: Url,
         chain_id: ChainId,
     ) -> Self {
         let geth_client = Client::new(url);
-        let dai_contract_adr = Address::random();
 
+        // In tests, the DAI contract doesn't exist until after _we_
+        // deploy it. We will replace this value once that happens
+        let random_dai_contract_address = Address::random();
```
D4nte, line 61:
```suggestion
        let random_dai_contract_address = Address::default();
```
Setting a placeholder to `0x0..0` makes more sense to me.

src/mid_market_rate.rs
luckysori: What do you propose instead? And what's the harm in using floating point arithmetic here? I'm a computer maths noob.

src/ethereum/wallet.rs
luckysori: Oh, nice point!

src/ethereum/wallet.rs
luckysori: I agree with this. I thought about doing it properly like you propose, but it was late and I was tired so I just re-exported it haha. I'll try!

src/ethereum/dai.rs
luckysori: Yeah, I'm pretty happy with it. We could even go a little further and make this return an error if you pass a public chain ID with the wrong DAI contract address. But this seems fine anyway.

src/mid_market_rate.rs
D4nte: The result might not be the one to expect if it cannot be stored in 64 (binary) bytes

luckysori: But wouldn't it just be a matter of losing a little bit of precision? In any case, I'm not too sure how I would go about avoiding this...

luckysori: > In any case, I'm not too sure how I would go about avoiding this...

I could try to use a dependency such as https://docs.rs/fixed/1.1.0/fixed/ if you're okay with it.

da-kami: I am fine with introducing a fixed-point library, I think in the long run it would probably make our life easier. 
Might be a good point to to try the library!

Note: I remember that there was quite some discussion around "use a library and trust it" vs. "write it ourselves to trust it" - alternatively we can also introduce another function for handling this division into `float_maths` (similar to `divide_pow_ten_trunc`).

luckysori: Okay, thanks for the feedback, I'll handle it in a separate PR :wink:

src/command/withdraw.rs
```diff
@@ -19,7 +19,7 @@ pub async fn withdraw(
         }
         Withdraw::Dai { amount, to_address } => {
             let tx_id = ethereum_wallet
-                .transfer_dai(to_address, amount.clone(), ethereum_wallet.chain_id)
+                .transfer_dai(to_address, amount.clone(), ethereum_wallet.chain_id())
```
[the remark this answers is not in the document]
luckysori: This one is a bit tricky. We normally use this argument to check if we're connected to the chain (or network for Bitcoin) we expect to be connected to. But withdrawing does not specify one, so we "cheat". I think this check is useful, so the proper fix would be to let the withdraw parameters determine the network/chain ID.

Keep in mind that wallets already knew about chain ID/network before this PR. What does @D4nte reckon?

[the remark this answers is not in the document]
D4nte: I was lazy and re-passed the `ChainId` here. Where is `transfer_dai` used? is it only used here, if so then we should remove the argument. If it is used both for a comit action and here then maybe we could make it an `Option`?

PR #92, opened 2020-07-30
Improve config documentation

D4nte:
Use a sample toml config file instead of embedded in markdown.
Parse and check the result as part of the tests.


D4nte:
bors retry

PR #93, opened 2020-07-31
Resume only

D4nte:
New command `nectar wrap-up` that finishes executing ongoing swaps.
- It does not publish new swaps
- It rejects any new take request
- Exists once all swaps are finished

UX would be:
```
nectar trade
# CTRL-C
nectar resume-only
```

Patches can be reviewed individually.


src/command/trade.rs
```diff
@@ -352,37 +343,13 @@ fn handle_network_event(
     ethereum_wallet: Arc<ethereum::Wallet>,
     bitcoin_connector: Arc<comit::btsieve::bitcoin::BitcoindConnector>,
     ethereum_connector: Arc<comit::btsieve::ethereum::Web3Connector>,
+    take_request_handler: F,
```
bonomat, line 346: Why do you pass in the logic for processing the take request as a function from the outside?

bonomat:
Thanks for asking me for a review. I don't know anything about the codebase so will leave the big comments to others :) 

I've added one question out of interest though.

src/command/trade.rs
D4nte: Because depending on whether I am in `trade` or `wrap-up` then I would process the take request differently.

D4nte:
> Thanks for asking me for a review. I don't know anything about the codebase so will leave the big comments to others :)
> 
> I've added one question out of interest though.

I'd like some UX input from you and @thomaseizinger 😉

bonomat:
> > Thanks for asking me for a review. I don't know anything about the codebase so will leave the big comments to others :)
> > I've added one question out of interest though.
> 
> I'd like some UX input from you and @thomaseizinger 😉

```bash
nectar trade
# CTRL-C
nectar wrap-up
```

I think that would do the job 👍 
Maybe printing on the console how many swaps are pending would be a plus.

src/command/resume_only.rs
```diff
@@ -0,0 +1,131 @@
...
+            Arc::clone(&db),
+            Arc::clone(&bitcoin_wallet),
+            Arc::clone(&ethereum_wallet),
+            Arc::clone(&bitcoin_connector),
+            Arc::clone(&ethereum_connector),
+            swap,
+        )
+        .await?;
+        handle_finished_swap(finished_swap, Arc::clone(&db), Arc::clone(&history))
+    }
```
luckysori, line 65: :mega: I think we should be joining all these futures rather than awaiting one by one.

D4nte: Done.

PR #94, opened 2020-07-31
Actually build on bors' branches

D4nte:
😑


thomaseizinger:
Uhm, so you were using bors but also not really? :D

D4nte:
> Uhm, so you were using bors but also not really? :D

It's more: "no wonder bors always timed out".

PR #95, opened 2020-07-31
Fix default-config loading bug.

bonomat:
I found a small bug when loading the config from the default location.

`ok_or_else` was used which essentially mapped to `Result<PathBuf, Error(Result<Path, Error>))`

Which resulted in the following behavior: 
if nectar was started with `-c /default/path/to/config/"` it would return Ok("/default/path/to/config/"). 
However, if the config param was left away, the mapping function would result in `Err(Ok("/default/path/to/config/")). 
At the end you were calling `ok()` on that result which mapped `Err(...)` to `Option::None` and hence always defaulting to the default config. 

Sorry for not providing a test and maybe there is also a neater way to solve this than using `map_or_else` which would get rid of the `to_path` mapping :)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -46,23 +46,25 @@ pub fn read_config(config_file: &Option<PathBuf>) -> anyhow::Result<File> {
             eprintln!("Using config file {}", path.display());
             path
         })
-        .ok_or_else(|| {
-            // try to load default config
-            let default_path = crate::fs::default_config_path()?;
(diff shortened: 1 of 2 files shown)


D4nte:
That's fresh new code from #92... Can you add a test? :D (or I can add one later).

D4nte:
I'll take over and add a test.

D4nte:
@bonomat see the latest changes :_

PR #98, opened 2020-08-03
Simplify swap execution

luckysori:
Separates `hbit_herc20` into `hbit_herc20_alice` and `hbit_herc20_bob` (same for `herc20_hbit`). 

The value of having a single function for both parties (less code to maintain) was outweighed by the difficulty of modelling executing an action and watching for the event produced by the execution of an action through the same function/trait. These new functions can differentiate between acting and watching, which makes them easier to understand and makes it easier for us to handle edge cases e.g. not having to wait for the other party to refund if we have aborted the swap execution.

I've tried to separate out things that could be upstreamed to COMIT lib into the module `swap/comit`. Interestingly, the function `try_do_it_once` stays within Nectar. The idea is that instead of using this function directly within the swap execution functions, implementations of `{hbit, herc20}::ExecuteACTION` traits in Nectar use it to ensure that swap execution is idempotent and that performing an action for the first time fails if Beta has expired. Other applications may end up needing similar logic, but the previous model was too complex and abusing traits.


D4nte:
The `try_do_it_once` problematic is being addressed with one of the upcoming projects in cnd.

src/swap.rs
```diff
@@ -278,18 +120,43 @@ pub struct SwapParams {
     pub taker: Taker,
 }
 
+/// Fetch the current `Timestamp` for the a ledger.
```
D4nte, line 123: There is a no such thing as a "timestamp" for a ledger. It'd be good to clarify what is expected. For example, it could the current time in relation of the expiry functionality of the COMIT HTLCs.

src/swap/comit.rs
```diff
@@ -0,0 +1,346 @@
...
+        }
+    };
+
+    let hbit_redeem = hbit::watch_for_redeemed(
+        bitcoin_connector,
+        &hbit_params.shared,
+        hbit_funded.location,
+        start_of_swap,
+    );
+    let hbit_refund = alice.execute_refund(hbit_params, hbit_funded);
```
thomaseizinger, line 85: Alice already redeemed at this stage no (line 68)? Not really a point in executing refund :)

src/swap/comit.rs
```diff
@@ -0,0 +1,346 @@
...
+        + BlockByHash<Block = ::bitcoin::Block, BlockHash = ::bitcoin::BlockHash>,
+    EC: LatestBlock<Block = ethereum::Block>
+        + BlockByHash<Block = ethereum::Block, BlockHash = ethereum::Hash>
+        + btsieve::ethereum::ReceiptByHash,
+{
+    let hbit_funded =
+        match hbit::watch_for_funded(bitcoin_connector, &hbit_params.shared, start_of_swap).await {
+            Ok(hbit_funded) => hbit_funded,
+            Err(_) => return Ok(()),
+        };
```
thomaseizinger, line 122: This is exactly the code pattern that `?` expands to except that `?` returns an error. What do you think of this:

```suggestion
    let hbit_funded = hbit::watch_for_funded(bitcoin_connector, &hbit_params.shared, start_of_swap).await.context("hbit not funded")?;
```

src/swap/comit.rs
```diff
@@ -0,0 +1,346 @@
...
+    .await
+    {
+        Ok(herc20_redeemed) => herc20_redeemed,
+        Err(_) => {
+            bob.execute_refund(herc20_params, herc20_deployed, start_of_swap)
+                .await?;
+
+            return Ok(());
+        }
+    };
```
thomaseizinger, line 155: The transformation to `?` doesn't work out of the box here because we want to "recover" in the case of an error. However, we could have a happy path function that simply does `?` on every step and then this function calls the happy path one, checks for the error and "recovers" with `bob.execute_refund` in certain cases :)

src/swap/comit.rs
luckysori: Given that this is in the body of `hbit_herc20_alice`, I think she should always try to refund after she's done redeeming the ERC20 tokens. If anything to make sure that someone claims the money in case Bob fails to.

On the other hand, I think we should actually not care whether Bob redeems or not. From Alice's perspective at this stage she only cares about maybe being able to refund for free.

In conclusion, for swap executions involving Alice, I'm just going to call `alice.execute_refund` at the end.

src/swap/comit.rs
luckysori: That's a great idea!

src/swap/comit.rs
luckysori: Good idea!

src/swap/comit.rs
luckysori: > In conclusion, for swap executions involving Alice, I'm just going to call alice.execute_refund at the end.

In practice this doesn't work so well, because we may end up waiting forever for this possible refund. In Nectar we rely on knowing when the swap execution ends to free funds and save the swap as a trade to a history file.

This could be solved by simply spawning the refund task away, but this would add an implicit (or explicit, if we decided to pass a handle as an argument) dependency on `tokio::Runtime`.

Thoughts, @thomaseizinger?

thomaseizinger: Though questions :D

It really boils down to application policy I guess? Are you gonna try and claim money that actually belongs to the other party? I am tempted to say we should just wait for the other party to redeem. In the end, the coins will either go to Alice or Bob, none else can take them so there isn't really a risk of "OMG free money someone's gotta take it :D"

> This could be solved by simply spawning the refund task away, but this would add an implicit (or explicit, if we decided to pass a handle as an argument) dependency on tokio::Runtime.

Alternatively, you could return the future from this function :)

luckysori: > I am tempted to say we should just wait for the other party to redeem. In the end, the coins will either go to Alice or Bob, none else can take them so there isn't really a risk of "OMG free money someone's gotta take it :D"

Yeah, I guess the library shouldn't enforce taking the other party's money if they mess up. Instead of waiting for Bob to redeem though, I would just consider the protocol done from Alice's perspective and exit.

luckysori:
@thomaseizinger One last review request so that you can see how I've implemented your suggestions.

PR #100, opened 2020-08-03
Database improvements

D4nte:
Few things on the DB:
- Quickcheck tests
- Store values in CBOR instead of JSON: The main reason is that it's a binary format because the DB stores byte arrays. It made less sense to use JSON than CBOR but there is little different considering the usecase.
- flush the db when storing a swap (arguably the most import patch of the PR)

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -631,3 +631,22 @@ mod transport {
         Ok(transport)
     }
 }
+
+#[cfg(test)]
+mod arbitrary {
+    use super::*;
+    use libp2p::multihash;
+    use quickcheck::{Arbitrary, Gen};
+
+    impl Arbitrary for Taker {
+        fn arbitrary<G: Gen>(g: &mut G) -> Self {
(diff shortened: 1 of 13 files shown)


src/network.rs
```diff
@@ -632,3 +632,22 @@ mod transport {
...
     }
 }
+
+#[cfg(test)]
+mod arbitrary {
+    use super::*;
+    use libp2p::multihash;
+    use quickcheck::{Arbitrary, Gen};
+
+    impl Arbitrary for Taker {
```
thomaseizinger, line 642: I can recommend looking into proptest as an alternative to quickcheck: https://docs.rs/proptest/0.10.0/proptest/

To nice thing is, it doesn't really on traits being implemented, you are just composing functions together.

D4nte: proptest does not support async test, it's why I opted for quickcheck.

thomaseizinger: ah bummer :/

PR #101, opened 2020-08-03
Let tokio chose correct runtime::Handle to fix libp2p connection problem

da-kami:
Decided to put it in a separate PR to get it in fast, the Draft for the orderbook upgrade might still need some iterations.

Thanks to @luckysori @thomaseizinger !

### DIFF
--- a/src/command/trade.rs
+++ b/src/command/trade.rs
@@ -23,7 +23,6 @@ use std::{sync::Arc, time::Duration};
 const ENSURED_CONSUME_ZERO_BUFFER: usize = 0;
 
 pub async fn trade(
-    runtime_handle: tokio::runtime::Handle,
     seed: &Seed,
     settings: Settings,
     bitcoin_wallet: bitcoin::Wallet,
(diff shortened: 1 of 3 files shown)


da-kami:
> Shall we have a test? :D

What and how would you test it? 

We could test the network layer, but I am not sure we could have caught this problem with network tests. That might require end2end tests that actually run `main`. I think we should have end2end tests that run the application at some point, the question is just when do we invest the time. I will focus on the app for a while, my objective was to "get Nectar working with the app".

da-kami:
> Does this fix things? Nice :)
Yes :)

> Might as well just pass a closure and do tokio::spawn now.
> That would make things even simpler because you don't need the TokioExecutor. I think there is an executor_fn method.

I had a look and did not see a function that would allow that. Feel free to push a commit if you want :)

D4nte:
Yes sorry. I thought more about it, you can't really test that. The only way would be to have an integration test that setup things the same way than the main and simulate a taker but it's a lot of overhead.

thomaseizinger:
> I had a look and did not see a function that would allow that. Feel free to push a commit if you want :)

Just realized that [this one](https://docs.rs/libp2p-core/0.20.1/libp2p_core/network/struct.NetworkConfig.html#method.set_executor_fn) is not exposed in libp2p for some reason, hmm...

thomaseizinger:
@da-kami there you go: https://github.com/libp2p/rust-libp2p/pull/1686 :)

PR #103, opened 2020-08-07
Add decision which swap to spawn based on position to production code

da-kami:
Somehow we overlooked changing the code in the production code 😅
When testing the execution together with the TakerUI we ran into the problem that Nectar only spawns `HbitHerc20` swaps.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -417,14 +417,24 @@ impl NetworkBehaviourEventProcess<network::comit::BehaviourOutEvent> for Nectar 
                     chain_id: order.ethereum_ledger.chain_id,
                 };
 
-                let swap = SwapKind::HbitHerc20(SwapParams {
-                    hbit_params,
-                    herc20_params,
-                    secret_hash,
-                    start_of_swap: Utc::now().naive_local(),
-                    swap_id: crate::SwapId::default(),
-                    taker,
-                });
+                let swap = match order.position {
+                    comit::network::Position::Buy => SwapKind::Herc20Hbit(SwapParams {
+                        hbit_params,
+                        herc20_params,
+                        secret_hash,
+                        start_of_swap: Utc::now().naive_local(),
+                        swap_id: crate::SwapId::default(),
+                        taker,
+                    }),
+                    comit::network::Position::Sell => SwapKind::HbitHerc20(SwapParams {
+                        hbit_params,
+                        herc20_params,
+                        secret_hash,
+                        start_of_swap: Utc::now().naive_local(),
+                        swap_id: crate::SwapId::default(),
+                        taker,
+                    }),
+                };
 
                 self.events.push_back(Event::SpawnSwap(swap));
             }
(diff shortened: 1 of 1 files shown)


src/network.rs
```diff
@@ -417,14 +417,24 @@ impl NetworkBehaviourEventProcess<network::comit::BehaviourOutEvent> for Nectar
...
-                let swap = SwapKind::HbitHerc20(SwapParams {
-                    hbit_params,
-                    herc20_params,
-                    secret_hash,
-                    start_of_swap: Utc::now().naive_local(),
-                    swap_id: crate::SwapId::default(),
-                    taker,
-                });
+                let swap = match order.position {
+                    comit::network::Position::Buy => SwapKind::Herc20Hbit(SwapParams {
```
rishflab, line 421: I think Buy becomes HbitHerc20 and Sell becomes Herc20Hbit

thomaseizinger: Unfortunately, we have to flip this because `buy` and `sell` are relative to the actor. The same swap is a `buy` order for Alice and a `sell` order for Bob.

This kind of confusion was the primary reason we invented the alpha/beta terminology :)

Nectar for the moment always takes the role of `Bob` whereas most of cnd's orderbook API takes the role of `Alice`. We should probably unify this code somewhere in the comit lib to avoid that every client has to get this right themselves :sweat_smile: 

cc @tcharding

da-kami: I think you are right! Good catch.

Reasoning:
If we are Bob we are on the `beta` ledger, hence `Herc20Hbit` would mean that we are funding `bitcoin`. If it is a `Buy` order, from the maker's perspective, we are buying `bitcoin` (the base), and should hence fund `erc20`, not `bitcoin`. So it should be flipped. @thomaseizinger I remember that we had a look and applied it like this - is it actually correct on cnd's side if I flip it? 😅
Let's have a look together on Monday. Note: I flipped it in the setup and, weirdly, Nectar does not start watching then.

tcharding: > Unfortunately, we have to flip this because buy and sell are relative to the actor. The same swap is a buy order for Alice and a sell order for Bob.

Yes, in any transaction one party sells something and the other party buys it. When an order is created (i.e., by maker) it is also commonly referred to as a bid or ask (offer). The terms bid/ask only ever apply to the maker, never the taker - for this reason they are useful terms. The mapping from bid/ask to buy/sell can be remembered by noticing that bid and buy start with a 'b' and sell and ask both contain an 's'. An exchange will often name the orderbook with the list of 'buys' being all the current bids and 'sells' being all the current asks - this is again from the perspective of the maker.

Mapping from forex terminology to COMIT terminology is hard and confusing, I have not come up with any mnemonics to help, anyone else?

thomaseizinger: > The terms bid/ask only ever apply to the maker, never the taker

Which is the confusing part that annoys the shit out of me :joy:

Can someone write up a concise cheat-sheet that translates between those two worlds please? :D

thomaseizinger: @da-kami 

I thought we did it correctly but we could be wrong. Might also be that `cnd` is wrong :man_shrugging:

da-kami:
The code is correct now (this code corresponds to the `new-orderbook` branch where we did a successful swap).

PR #104, opened 2020-08-07
Wallet rescan to pick up funding that happened prior to creating the wallet

da-kami:
When funding the deposit address prior to creating the wallet the funding is not picked up. 
Adding `rescan` fixes this.

### DIFF
--- a/src/bitcoin/wallet.rs
+++ b/src/bitcoin/wallet.rs
@@ -52,26 +52,29 @@ impl Wallet {
                     .create_wallet(&self.name, None, Some(true), None, None)
                     .await?;
 
-                self.bitcoind_client.rescan(&self.name).await?;
-
                 let wif = self.seed_as_wif();
 
                 self.bitcoind_client
                     .set_hd_seed(&self.name, Some(true), Some(wif))
-                    .await
+                    .await?;
+
+                self.bitcoind_client.rescan(&self.name).await?;
+
+                Ok(())
             }
             Ok(WalletInfoResponse {
                 hd_seed_id: None, ..
             }) => {
-                // Rescan wallet to ensure funding is picked up
-                self.bitcoind_client.rescan(&self.name).await?;
-
                 // The wallet may have been previously created, but the `sethdseed` call may have failed
                 let wif = self.seed_as_wif();
 
                 self.bitcoind_client
                     .set_hd_seed(&self.name, Some(true), Some(wif))
-                    .await
+                    .await?;
+
+                // Rescan wallet to ensure funding is picked up
+                self.bitcoind_client.rescan(&self.name).await?;
+                Ok(())
             }
             _ => {
                 // Rescan wallet to ensure funding is picked up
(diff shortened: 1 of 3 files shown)


src/bitcoin/wallet.rs
```diff
@@ -52,6 +52,8 @@ impl Wallet {
                     .create_wallet(&self.name, None, Some(true), None, None)
                     .await?;
 
+                self.bitcoind_client.rescan(&self.name).await?;
```
D4nte, line 55: 📣 The `rescan` must be done **after** the seed is set. At this point in time, there is no key to check against.

src/bitcoin/wallet.rs
```diff
@@ -61,14 +63,21 @@ impl Wallet {
             Ok(WalletInfoResponse {
                 hd_seed_id: None, ..
             }) => {
+                // Rescan wallet to ensure funding is picked up
+                self.bitcoind_client.rescan(&self.name).await?;
```
D4nte, line 67: `rescan` **after** the seed is set.

src/bitcoin/wallet.rs
```diff
@@ -61,14 +63,21 @@ impl Wallet {
...
                 let wif = self.seed_as_wif();
 
                 self.bitcoind_client
                     .set_hd_seed(&self.name, Some(true), Some(wif))
                     .await
             }
-            _ => Ok(()),
+            _ => {
+                // Rescan wallet to ensure funding is picked up
+                self.bitcoind_client.rescan(&self.name).await?;
```
D4nte, line 78: This is unnecessary as the wallet would be created on bitcoind already.

src/test_harness/bitcoin.rs
```diff
@@ -40,6 +40,8 @@ impl<'c> Blockchain<'c> {
             .create_wallet(&self.wallet_name, None, None, None, None)
             .await?;
 
+        bitcoind_client.rescan(&self.wallet_name).await?;
```
D4nte, line 43: How can there be founding a new random wallet? This is useless.

src/bitcoin/wallet.rs
da-kami: This is necessary, because otherwise we don't pick up money if the address was funded prior to the wallet being created.
This can happen if I run Nectar to e.g. display the deposit addresses (creates the wallet...) and then fund the address and then run it again against the same node.

src/test_harness/bitcoin.rs
da-kami: removed.

src/bitcoin/wallet.rs
D4nte: This matching branch means the wallet was already created on bitcoind. It means nectar is **not** creating the wallet, the wallet was already here.

Please describe a sequence of event that would justify re-scanning here.

If the `rescan` is properly done in the other match branches (when the wallet **is** created by nectar) then it should be enough.

da-kami: You are right, thanks! Somehow I thought I remember that I needed it - but that was probably some weird scenario when testing the new setup.
I just tried it with the current comit-script setup that pre-funds (and then Nectar is started, creates the wallet, rescans - upon starting it again the rescan is not needed).

Removed it!

PR #110, opened 2020-08-25
Only insert swap in db if we are spawning it for the first time

### LINKED ISSUE #107: Nectar does not properly restart the swap execution after `respawn`ing an unfinished swap
Setup:

1. Take `SELL` order in taker UI
2. Swap started between taker `cnd` and nectar.
3. `cnd` redeems Bitcoin and yields `swap finished`
4. Nectar gets stuck looking for `cnd`'s redeem.
5. Kill nectar.
6. After restarting Nectar it does not properly trigger swap execution in respawn.

Quick investigation yielded, that this statement is causing problems:

Respawn triggers swap execution:

https://github.com/coblox/nectar/blob/e873a482c0242e5de32e9cfb46cf60e0b81a622b/src/command/trade.rs#L259-L291

But we fail when calling the insert statement on the DB. The actual execution is never triggered. 

This is most probably because we run into the `Ok` path when loading the swap.

https://github.com/coblox/nectar/blob/713199dafe14ae9008aede40a8024bceb90f75c6/src/swap/db.rs#L69-L95

However, I never see the error `anyhow!("Swap is already stored")` - how is this lost? Sync db vs async problem?

Did not investigate further from there.

D4nte:
Resolves #107.

It was actually a simple functional bug.

The testing seems quite heavy and the bug seemed obvious when reasoning about it so no test added.

### DIFF
--- a/src/command/resume_only.rs
+++ b/src/command/resume_only.rs
@@ -80,8 +80,6 @@ async fn execute_swap(
     ethereum_connector: Arc<comit::btsieve::ethereum::Web3Connector>,
     swap: SwapKind,
 ) -> anyhow::Result<FinishedSwap> {
-    db.insert_swap(swap.clone()).await?;
-
     swap.execute(
         Arc::clone(&db),
         Arc::clone(&bitcoin_wallet),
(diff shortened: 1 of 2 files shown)


luckysori (approved):
Nice.

PR #111, opened 2020-08-25
Fix nectar not seeing redeem

### LINKED ISSUE #108: Investigate nectar missing swap events (that are properly detected on cnd's side)
Setup:

1. Take SELL order in taker UI
2. Swap started between taker cnd and nectar.
3. cnd redeems Bitcoin and yields swap finished
4. Nectar gets stuck looking for cnd's redeem.

The redeem is properly detected on on cnd's side (yielded `swap finished`). Nectar kept looking for the transaction.

I noticed this behaviour on multiple occasions. My hunch, that Nectar somehow does not properly initializes the timestamp for watching transactions for btsieve, could not be proven to be the problem, we seem to set that correctly:

https://github.com/coblox/nectar/blob/e873a482c0242e5de32e9cfb46cf60e0b81a622b/src/network.rs#L449-L529

It could be, that we somehow set the timestamp different when starting to look for redeem?

Did not investigate further.

D4nte:
Resolves #108.


thomaseizinger:
Thank you for investigating and fixing this <3

src/network.rs
```diff
@@ -426,7 +426,7 @@ impl NetworkBehaviourEventProcess<network::comit::BehaviourOutEvent> for Nectar
                             hbit_params,
                             herc20_params,
                             secret_hash,
-                            start_of_swap: Utc::now().naive_local(),
+                            utc_start_of_swap: Utc::now().naive_local(),
```
luckysori, line 429: :mega: I think it's
```suggestion
                            utc_start_of_swap: Utc::now().naive_utc(),
```

luckysori (approved):
Nice! I think you missed one

luckysori (changes requested):
I meant to request changes oops

src/swap.rs
```diff
@@ -196,7 +196,7 @@ impl crate::StaticStub for SwapParams {
                 chain_id: 42.into(),
             },
             secret_hash: SecretHash::new(comit::Secret::from(*b"hello world, you are beautiful!!")),
-            start_of_swap: chrono::Local::now().naive_local(),
+            utc_start_of_swap: chrono::Local::now().naive_local(),
```
D4nte, line 199: And here!

src/swap/db.rs
```diff
@@ -247,7 +247,7 @@ impl StaticStub for Swap {
                 .unwrap(),
             ),
             taker: network::Taker::static_stub(),
-            start_of_swap: chrono::Local::now().naive_local(),
+            utc_start_of_swap: chrono::Local::now().naive_local(),
```
D4nte, line 250: and here

luckysori (approved):

PR #112, opened 2020-08-29
Order matching

rishflab:
Upgraded nectar to work with the new order matching engine and setup swap protocol in comit-rs. These changes are required for project tantalus.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ async-trait = "0.1"
 bitcoin = { version = "0.23.0", features = ["rand", "use-serde"] }
 chrono = "0.4"
 clarity = "0.1"
-comit = { git = "https://github.com/comit-network/comit-rs", package = "comit", rev = "2a5020205dec357fbf85d16bec08864ac9c8af80" }
+comit = { git = "https://github.com/comit-network/comit-rs/", package = "comit", rev = "cc450c425b52e060371d0384a759cd0eab13396c" }
 config = "0.10"
 conquer-once = "0.2"
 csv = "1.1"
@@ -21,7 +21,7 @@ ethereum-types = "0.9.0"
 futures = "0.3.5"
 futures-timer = "3.0.2"
 hex = "0.4"
-libp2p = { version = "0.21", default-features = false, features = ["tcp-tokio", "secio", "yamux", "mplex", "dns", "gossipsub", "request-response"] }
+libp2p = { version = "0.24", default-features = false, features = ["tcp-tokio", "secio", "yamux", "mplex", "dns"] }
 log = "0.4"
 num = "0.3"
 num256 = "0.2"
@@ -37,9 +37,10 @@ structopt = "0.3"
 strum = "0.18"
 strum_macros = "0.18"
 thiserror = "1.0"
+time = { version = "0.2", features = ["serde"] }
 tokio = { version = "0.2", features = ["macros", "time"] }
 toml = "0.5"
-tracing = "0.1.15"
+tracing = "0.1"
 tracing-log = "0.1"
 tracing-subscriber = "0.2"
 url = { version = "2", features = ["serde"] }
--- a/src/bitcoin/wallet.rs
+++ b/src/bitcoin/wallet.rs
@@ -254,16 +254,22 @@ impl Wallet {
             hash[0], hash[1], hash[2], hash[3]
         )
     }
+
+    pub fn seed(&self) -> Seed {
+        self.seed
+    }
 }
 
 #[cfg(all(test, feature = "test-docker"))]
 mod docker_tests {
     use super::*;
     use crate::test_harness::bitcoin;
-    use std::fs::File;
-    use std::io::{BufRead, BufReader};
-    use std::path::Path;
-    use std::process::{Command, Stdio};
+    use std::{
+        fs::File,
+        io::{BufRead, BufReader},
+        path::Path,
+        process::{Command, Stdio},
+    };
     use tempdir::TempDir;
     use testcontainers::clients;
 
--- a/src/command/resume_only.rs
+++ b/src/command/resume_only.rs
@@ -124,6 +124,10 @@ fn handle_finished_swap(
     let swap_id = finished_swap.swap.swap_id();
 
     let _ = db
+        .remove_active_peer(&finished_swap.peer)
+        .map_err(|error| tracing::error!("Unable to remove from active peers: {}", error));
+
+    let _ = db
         .remove_swap(&swap_id)
         .map_err(|error| tracing::error!("Unable to delete swap from db: {}", error));
 }
--- a/src/command/resume_only.rs
+++ b/src/command/resume_only.rs
@@ -103,7 +103,7 @@ fn handle_finished_swap(
 ) {
     {
         let trade = into_history_trade(
-            finished_swap.taker.peer_id(),
+            finished_swap.peer.peer_id(),
             finished_swap.swap.clone(),
             #[cfg(not(test))]
             finished_swap.final_timestamp,
@@ -122,10 +122,6 @@ fn handle_finished_swap(
     }
 
     let swap_id = finished_swap.swap.swap_id();
-
-    let _ = db
-        .remove_active_taker(&finished_swap.taker)
-        .map_err(|error| tracing::error!("Unable to remove from active takers: {}", error));
 
     let _ = db
         .remove_swap(&swap_id)
--- a/src/command/trade.rs
+++ b/src/command/trade.rs
@@ -4,20 +4,22 @@ use crate::{
     config::Settings,
     ethereum::{self, dai},
     history::History,
-    maker::{PublishOrders, TakeRequestDecision},
+    maker::PublishOrders,
     mid_market_rate::get_btc_dai_mid_market_rate,
     network::{self, Swarm},
     swap::{Database, SwapKind, SwapParams},
     Maker, MidMarketRate, Seed, Spread,
 };
 use anyhow::Context;
-use chrono::Utc;
 use comit::btsieve::bitcoin::BitcoindConnector;
 use comit::btsieve::ethereum::Web3Connector;
 use futures::channel::mpsc::Sender;
 use futures::{channel::mpsc::Receiver, Future, FutureExt, SinkExt, StreamExt, TryFutureExt};
 use futures_timer::Delay;
 
+use crate::maker::TakeRequestDecision;
+use crate::network::ActivePeer;
+use comit::{Position, Role};
 use std::{sync::Arc, time::Duration};
 
 const ENSURED_CONSUME_ZERO_BUFFER: usize = 0;
@@ -44,21 +46,24 @@ pub async fn trade(
     #[cfg(test)]
     let db = Arc::new(Database::new_test()?);
 
-    let mut swarm = Swarm::new(&seed, &settings, Arc::clone(&db))?;
+    let mut swarm = Swarm::new(
+        &seed,
+        &settings,
+        Arc::clone(&bitcoin_wallet),
+        Arc::clone(&ethereum_wallet),
+        Arc::clone(&db),
+    )?;
 
     let initial_sell_order = maker
         .new_sell_order()
         .context("Could not generate sell order")?;
+
     let initial_buy_order = maker
         .new_buy_order()
         .context("Could not generate buy order")?;
 
-    swarm
-        .publish(initial_sell_order.into())
-        .context("Could not publish initial sell order")?;
-    swarm
-        .publish(initial_buy_order.into())
-        .context("Could not publish initial buy order")?;
+    swarm.publish(initial_sell_order.to_order_comit_order(maker.swap_protocol(Position::Buy))?);
+    swarm.publish(initial_buy_order.to_order_comit_order(maker.swap_protocol(Position::Sell))?);
 
     let update_interval = Duration::from_secs(15u64);
 
@@ -162,6 +167,8 @@ async fn init_maker(
         spread,
         settings.bitcoin.network,
         settings.ethereum.chain,
+        // todo: get from config
+        Role::Bob,
     ))
 }
 
@@ -260,6 +267,8 @@ async fn execute_swap(
     mut finished_swap_sender: Sender<FinishedSwap>,
     swap: SwapKind,
 ) -> anyhow::Result<()> {
+    db.insert_swap(swap.clone()).await?;
+
     swap.execute(
         Arc::clone(&db),
         Arc::clone(&bitcoin_wallet),
@@ -273,7 +282,7 @@ async fn execute_swap(
         .send(FinishedSwap::new(
             swap.clone(),
             swap.params().taker,
-            Utc::now(),
+            chrono::Utc::now(),
         ))
         .await
         .map_err(|_| {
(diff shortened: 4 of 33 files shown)


src/command/trade.rs
```diff
@@ -44,21 +44,26 @@ pub async fn trade(
...
+        Arc::clone(&db),
+        Arc::clone(&bitcoin_wallet),
+        Arc::clone(&ethereum_wallet),
+    )?;
 
     let initial_sell_order = maker
         .new_sell_order()
         .context("Could not generate sell order")?;
+
+    println!("{:?}", initial_sell_order);
```
bonomat, line 59: This can be removed right? (same below)

src/command/trade.rs
```diff
@@ -399,12 +339,9 @@ fn handle_rate_update(
...
                     new_buy_order,
                 })) => {
-                    let _ = swarm
-                        .publish(new_sell_order.into())
-                        .map_err(|e| tracing::error!("Failed to publish new sell order: {}", e));
-                    let _ = swarm
-                        .publish(new_buy_order.into())
-                        .map_err(|e| tracing::error!("Failed to publish new buy order: {}", e));
+                    swarm.clear_own_orders();
+                    let _ = swarm.publish(new_sell_order.to_order_form(maker.swap_protocol()));
```
bonomat, line 343: Why did you remove handling the error?

bonomat:
Hi Rishab, please add a few words on the status of this _draft PR_, e.g. what's missing in your opinion and what's the goal, anything in particular we should look at.

bonomat (commented):
I did not finish the review yet but added a few comments.

rishflab:
> Hi Rishab, please add a few words on the status of this _draft PR_, e.g. what's missing in your opinion and what's the goal, anything in particular we should look at.

Sorry about that. Added in the purpose to the description.

src/bitcoin/wallet.rs
```diff
@@ -19,7 +19,7 @@ pub struct Wallet {
     /// The wallet is named `nectar_x` with `x` being the first 4 bytes of the hash of the seed
     name: String,
     bitcoind_client: Client,
-    seed: Seed,
+    pub seed: Seed,
```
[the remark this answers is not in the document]
rishflab: added a getter function and made it private

Cargo.toml
```diff
@@ -10,7 +10,7 @@ async-trait = "0.1"
 bitcoin = { version = "0.23.0", features = ["rand", "use-serde"] }
 chrono = "0.4"
 clarity = "0.1"
-comit = { git = "https://github.com/comit-network/comit-rs", package = "comit", rev = "2a5020205dec357fbf85d16bec08864ac9c8af80" }
+comit = { git = "https://github.com/comit-network/comit-rs/", package = "comit", branch = "nectar-tantalus" }
```
rishflab, line 13: Ill get this pointing to the commit once this branch is merged

src/command/trade.rs
```diff
@@ -516,17 +458,75 @@ async fn handle_finished_swap(
...
+                        tracing::error!("Sending setup swap message yielded error: {}", e)
+                    }
+                    // todo: publish new order here?
+                    // What if i publish a new order here and the does go through?
+                }
+                Ok(TakeRequestDecision::InsufficientFunds) => tracing::info!("Insufficient funds"),
+                Ok(TakeRequestDecision::RateNotProfitable) => tracing::info!("Rate not profitable"),
+                Err(e) => tracing::error!("Processing taken order yielded error: {}", e),
+            };
+        }
```
thomaseizinger, line 502: I guess for now this is fine. In the long run though, the `TakeRequestDecision` has to go away. Nectar is not supposed to make a decision here whether or not it should move forward otherwise the whole idea of the orderbook is rendered invalid :sweat_smile:

src/network.rs
```diff
@@ -1,37 +1,35 @@
...
     pin::Pin,
     str::FromStr,
     task::{Context, Poll},
 };
 
-use crate::swap::Database;
+use crate::swap::SwapParams;
+
+use chrono::Utc;
+
```
thomaseizinger, line 22: Not sure what nectar policy is here but your imports are scattered all over and hence not formatted :)

src/network.rs
```diff
@@ -117,119 +131,83 @@ pub struct SwapMetadata {
...
-            shared,
-        }
-    }
-}
-
-impl From<SwapId> for crate::SwapId {
-    fn from(from: SwapId) -> Self {
-        from.inner
-    }
+// todo: move this from cnd to comit and import it from comit
```
thomaseizinger, line 134: This should not be needed in nectar (is this dead code maybe?) and it is already gone from cnd as well :)

src/network.rs
```diff
@@ -117,119 +131,83 @@ pub struct SwapMetadata {
...
+    _configured_bitcoin_network: bitcoin::Network,
+    /// The Ethereum network we are currently connected to.
     #[behaviour(ignore)]
-    takers: HashMap<OrderId, PeerId>,
+    _configured_ethereum_network: ethereum::ChainId,
+    /// The address of the DAI ERC20 token contract on the current Ethereum
+    /// network.
     #[behaviour(ignore)]
-    database: Arc<Database>,
+    _dai_contract_address: ethereum::Address,
```
thomaseizinger, line 159: I reckon these could easily be features of the wallets that you already have in here.

If you are actually using them, you probably also want to remove the underscore :sweat_smile:

src/network.rs
```diff
@@ -246,232 +224,286 @@ impl Nectar {
...
+                                    erc20: comit::asset::Erc20 {
+                                        token_contract,
+                                        quantity: erc20_quantity,
+                                    },
+                                    bitcoin: quantity,
+                                    ethereum_absolute_expiry,
+                                    bitcoin_absolute_expiry,
+                                    ethereum_chain_id: self._configured_ethereum_network,
+                                    bitcoin_network: self._configured_bitcoin_network,
+                                },
```
thomaseizinger, line 301: CommonParams are exactly the same in both cases no? In fact, I think they should be the same for all combinations given that the params themselves don't say anything about the swap protocol.

src/network.rs
```diff
@@ -246,232 +224,286 @@ impl Nectar {
...
-                            hbit::SharedParams {
-                                asset: order.bitcoin_amount,
-                                redeem_identity,
-                                refund_identity: taker_bitcoin_identity,
-                                expiry: order.bitcoin_absolute_expiry.into(),
-                                secret_hash,
-                                network: order.bitcoin_ledger.into(),
+            ::comit::network::setup_swap::BehaviourOutEvent::ExecutableSwap(exec_swap) => {
+                let swap_id = exec_swap.context;
+                // todo: get from order match ref point
```
thomaseizinger, line 405: You can solve this TODO easily by introducing a context that contains the reference point. I did the same thing in cnd.

src/order.rs
```diff
@@ -96,8 +113,13 @@ impl BtcDaiOrder {
             None => quote_balance - quote_reserved_funds,
         };
 
+        println!("quote_amount: {:?}", quote_amount);
```
thomaseizinger, line 116: Can you remove those either entirely or replace with logs please?

src/swap.rs
```diff
@@ -118,7 +118,8 @@ pub struct SwapParams {
     pub hbit_params: hbit::Params,
     pub herc20_params: herc20::Params,
     pub secret_hash: comit::SecretHash,
-    pub utc_start_of_swap: NaiveDateTime,
+    // TODO: Why naive and not DateTime<Local>?
```
thomaseizinger, line 121: Seems like an outdated comment :)

src/swap.rs
```diff
@@ -240,10 +241,8 @@ mod arbitrary {
                 hbit_params: hbit::Params::arbitrary(g),
                 herc20_params,
                 secret_hash: secret_hash(g),
-                utc_start_of_swap: chrono::NaiveDateTime::from_timestamp(
-                    u32::arbitrary(g) as i64,
-                    0,
-                ),
+                // todo:   chrono::NaiveDateTime::from_timestamp(u32::arbitrary(g) as i64, 0);
+                start_of_swap: chrono::Utc::now(),
```
thomaseizinger, line 245: That is a bad idea as it will completely mess up the "[shrinking](https://github.com/BurntSushi/quickcheck#shrinking)" of quickcheck. You should be able to just get an `i64` from the `Gen` trait?

src/swap/db.rs
```diff
@@ -139,61 +137,6 @@ impl Database {
     }
 }
 
-/// Active takers related functions
-impl Database {
```
thomaseizinger, line 143: Are we sure we can just delete this functionality? I think nectar had a requirement of not doing more than one trade with a single person at the same time.

src/maker.rs
```diff
@@ -120,9 +117,17 @@ impl Maker {
         self.dai_balance = None;
     }
 
-    pub fn new_sell_order(&self) -> anyhow::Result<BtcDaiOrder> {
+    //todo: get proper expiries
```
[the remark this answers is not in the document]
thomaseizinger: We need to get this right otherwise we will never have a match between cnd and nectar.

src/network.rs
rishflab: there doesnt seem to be any auto formatting, I don't think it is worth the effort fixing this manually at the moment

src/swap/db.rs
rishflab: Im not sure to be honest. Why cant nectar do more than one trade at a time?

thomaseizinger: I think the reasoning was to minimize risk of trading with a bad counterparty. @d4nte do you mind chimming in here?

Cargo.toml
```diff
@@ -21,25 +21,27 @@ ethereum-types = "0.9.0"
...
 futures-timer = "3.0.2"
 hex = "0.4"
-libp2p = { version = "0.21", default-features = false, features = ["tcp-tokio", "secio", "yamux", "mplex", "dns", "gossipsub", "request-response"] }
+libp2p = { version = "0.24", default-features = false, features = ["tcp-tokio", "secio", "yamux", "mplex", "dns"] }
 log = "0.4"
 num = "0.3"
 num256 = "0.2"
 pem = "0.8"
-reqwest = { version = "0.10", default-features = false, features = ["json", "native-tls"] }
+reqwest = { version = "0.10.8", default-features = false, features = ["json", "native-tls"] }
```
D4nte, line 29: Can you explain this change? the lock file should sort that out for you.

Cargo.toml
```diff
@@ -21,25 +21,27 @@ ethereum-types = "0.9.0"
...
 serde = { version = "1", features = ["derive"] }
 serde-hex = "0.1"
 serde_cbor = "0.11"
 serde_json = "1.0"
+sha2 = "0.9"
```
D4nte: I don't think this is needed. Let me try to remove it.

Cargo.toml
```diff
@@ -21,25 +21,27 @@ ethereum-types = "0.9.0"
...
 spectral = "0.6"
 structopt = "0.3"
 strum = "0.18"
 strum_macros = "0.18"
 thiserror = "1.0"
+time = { version = "0.2.16", features = ["serde"] }
 tokio = { version = "0.2", features = ["macros", "time"] }
 toml = "0.5"
-tracing = "0.1.15"
+tracing = "0.1.19"
```
D4nte, line 43: Just remove the patch digits.

rustfmt.toml
```diff
@@ -0,0 +1,11 @@
+edition = "2018"
```
D4nte, line 1: 📣 Please introduce formatting rule in a dedicated PR where you also add a check in the CI to enforce those rules.

src/bitcoin/amount.rs
```diff
@@ -17,7 +17,8 @@ pub struct Asset {
 pub struct Amount(::bitcoin::Amount);
 
 impl Amount {
-    // The rate input is for bitcoin to dai but we applied to satoshis so we need to:
+    // The rate input is for bitcoin to dai but we applied to satoshis so we need
+    // to:
```
D4nte, line 21: This is noise, see my comment on the formatting rules.

src/bitcoin/wallet.rs
```diff
@@ -254,16 +259,22 @@ impl Wallet {
             hash[0], hash[1], hash[2], hash[3]
         )
     }
+
+    pub fn seed(&self) -> Seed {
```
D4nte, line 263: I see this is done to generate a secret hash.
It is strange because the seed is being taken from the Bitcoin Wallet, I now realize that the seed stays in memory which is not great for security concern.

If you need a `Seed` to generate the secret then pass it direcly where you need it instead of having this indirection `Seed -> BitcoinWallet -> Seed -> SecretHash`.

Maybe better to actually hash the seed and use the hash of the seed as a seed for the secret hash. The seed is the most important secret here so we should avoid passing it around.

Cargo.toml
D4nte: No need to introduce a new dependency, see ebb54d13.

src/command/resume_only.rs
```diff
@@ -123,10 +127,6 @@ fn handle_finished_swap(
 
     let swap_id = finished_swap.swap.swap_id();
 
-    let _ = db
-        .remove_active_taker(&finished_swap.taker)
```
D4nte, line 127: Can you explain please?

src/ethereum/dai.rs
```diff
@@ -132,7 +134,12 @@ impl Amount {
         // Apply the rate
         let (worth, _remainder) = self.as_atto().div_rem(&uint_rate);
 
+        if worth.is_zero() {
```
D4nte, line 137: You can add a proptest on this function to ensure there is no other arithmetic issues.

src/ethereum/wallet.rs
```diff
@@ -145,15 +145,16 @@ impl Wallet {
             .send_raw_transaction(transaction_hex)
             .await?;
 
+        #[allow(unreachable_patterns)]
```
D4nte, line 148: Please add a comment when adding `allow` attributes.
Also, the compiler is right, the last match branch is unreachable, just delete it.

D4nte (changes requested):
Please create a separate PR for the formatting, it is creating unecessary work for the reviewer.

I have not finished to review, will finish once formatting is out.
Please do not introduced `sha2` dependency, I pushed a patch to show how it can be done.

src/command/resume_only.rs
rishflab: I removed the code which saves active takers to the db so this wont do anything. The new order matching engine has no concept of Takers. I am not 100% how this affects nectar.

rustfmt.toml
rishflab: I think there already is a check in CI, I introduced this because it was failing the check

Cargo.toml
rishflab: I was getting dependency conflicts, I will try again

src/command/resume_only.rs
D4nte: Ok, allow me to give you a bit more context.

To mitigate DDOS type attack, Nectar does not execute 2 swaps with the same counterpart.
Hence, we store `takers` in the DB when starting a swap, and delete them when the swap is finished.

When a new swap is meant to start, we check if a swap is already ongoing with the given counterparty. If it is, then we do not start the swap.

I believe that this still fits in the new design. If 2 parties match orders, then it's ok to have a consensus rule to not start a new swap if they are already executing one.

I believe this feature is good to have and if we were to remove it, it should be by choice.

I'd be keen to keep the feature. @thomaseizinger @bonomat please share your thoughts,

rustfmt.toml
D4nte: If your PR fails the format check then just run `cargo fmt` locally, no need to introduce `rustfmt.toml` or use nightly.

src/swap/db.rs
D4nte: Done here: https://github.com/coblox/nectar/pull/112#discussion_r480506366

src/command/resume_only.rs
bonomat: I agree, it is still a good feature to keep. 
It does not give you a 100% guaranty that you don't get gamed but it makes it at least harder.

src/bitcoin/wallet.rs
rishflab: Agree with the hash idea. Is it ok if this is done in a seperate PR?
The indirection is there because I did not want to duplicate the seed in the Nectar struct. I had initially done it this way

src/command/resume_only.rs
thomaseizinger: Yep, it looks like we just might need to change some terminology at some point!

src/ethereum/dai.rs
rishflab: This will all get sorted when we change to price quantity

bonomat:
Hey Rishab.
As you might have noticed, you've removed the `rustfmt.toml` but the formatting changes remain. Instead of removing the `rustfmt.toml` file with your commit in [`379c8b6` (#112)](https://github.com/coblox/nectar/pull/112/commits/379c8b6e877b0678cbb20be6be52dba0d93190d7) you could do an interactive rebase back to [`24e36e0` (#112)](https://github.com/coblox/nectar/pull/112/commits/24e36e0d828b9b37587aff49abcb2ba96be6718d#diff-1a80e1fce76d9ad7a4abc1720d360380R3) and not add it in the first place. That way your history remains clean and some formatting changes will never happen (such as the one @D4nte commented on: https://github.com/coblox/nectar/pull/112/files#r480484323).

src/bitcoin/wallet.rs
D4nte: Yes, feel free to clean up in a separate PR. Adding a `TODO` here would be a good way to track a low hanging fruit clean up.

src/command/resume_only.rs
D4nte: Yes, feel free to change to `Counterpart` here or just add a TODO to rename later. Either way, I think the feature needs to be kept and I understand it follow the new matching consensus principle.

src/swap/db.rs
rishflab: re added these checks

src/command/resume_only.rs
rishflab: Added this functionality back in and renamed it

src/bitcoin/wallet.rs
rishflab: there is a todo in the Nectar struct which will link to this.

src/command/trade.rs
rishflab: I will add a todo

da-kami:
Sorry for not investigating further.
When trying to run this for the demo I noticed some issues:

1. The order that was published was incorrect by one decimal point (instead of 12000 it was 1200). I think this is either a precision or a general conversion error.

2. Once Nectar "re-publishes" (i.e. cancel + new order) orders cnd never picks them up. 
Since the initial orders are picked up this is most likely a problem of cancelling orders.

Could not investigate further so far.

rishflab:
> Sorry for not investigating further.
> When trying to run this for the demo I noticed some issues:
> 
> 1. The order that was published was incorrect by one decimal point (instead of 12000 it was 1200). I think this is either a precision or a general conversion error.
> 2. Once Nectar "re-publishes" (i.e. cancel + new order) orders cnd never picks them up.
>    Since the initial orders are picked up this is most likely a problem of cancelling orders.
> 
> Could not investigate further so far.

hmm thanks for the feedback, ill see what i can do tommorow morning.

da-kami:
> > Sorry for not investigating further.
> > When trying to run this for the demo I noticed some issues:
> > 
> > 1. The order that was published was incorrect by one decimal point (instead of 12000 it was 1200). I think this is either a precision or a general conversion error.
> > 2. Once Nectar "re-publishes" (i.e. cancel + new order) orders cnd never picks them up.
> >    Since the initial orders are picked up this is most likely a problem of cancelling orders.
> > 
> > Could not investigate further so far.
> 
> hmm thanks for the feedback, ill see what i can do tommorow morning.

No worries, we got a demo with 2 UIs working :)

src/command/resume_only.rs
D4nte: @rishflab: I see your renamed `Taker` to `Peer` but I think you missed the re-addition of the call here to remove from the db once a swap is finished here.

D4nte (changes requested):
Note you still remove the call to remove the active peer from the DB.
Except for that, LGTM. Please note my functional review is limited as I am fully aware of the final design you are aiming for.

rishflab:
> Note you still remove the call to remove the active peer from the DB.
> Except for that, LGTM. Please note my functional review is limited as I am fully aware of the final design you are aiming for.

Yep your right,

D4nte (approved):

thomaseizinger (approved):

PR #114, opened 2020-09-07
Introduce seed for network purposes

D4nte:
Instead of carrying around the root seed that is used to initialised
the crypto wallets, we used a `network::Seed` for network purposes.

The aim here is to protect the wallet seed from any bug in the network
module that would leak it.

It also removes the indirection of getting the seed via the Bitcoin
wallet.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -1,6 +1,4 @@
-use crate::{
-    bitcoin, ethereum, ethereum::dai, order::BtcDaiOrderForm, swap::SwapKind, Seed, SwapId,
-};
+use crate::{bitcoin, ethereum, ethereum::dai, order::BtcDaiOrderForm, swap::SwapKind, SwapId};
 use ::bitcoin::hashes::{sha256, Hash, HashEngine};
 
 use comit::{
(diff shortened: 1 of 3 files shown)


src/network.rs
```diff
@@ -610,15 +626,36 @@ impl<'de> Deserialize<'de> for ActivePeer {
...
+    }
 
-    let hash = sha256::Hash::from_engine(engine);
-    let key = ed25519::SecretKey::from_bytes(hash.into_inner()).expect("we always pass 32 bytes");
-    libp2p::identity::Keypair::Ed25519(key.into())
+    pub fn bytes(&self) -> [u8; SEED_LENGTH] {
+        self.0
+    }
+
+    pub fn derive_key_pair(&self) -> libp2p::identity::Keypair {
```
thomaseizinger, line 648: Call this `derive_libp2p_identity` maybe?

src/network.rs
```diff
@@ -151,6 +153,12 @@ pub struct Nectar {
     orderbook: orderbook::Orderbook,
     setup_swap: setup_swap::SetupSwap<SetupSwapContext>,
     #[behaviour(ignore)]
+    seed: Seed,
+    #[behaviour(ignore)]
+    key_pair: Keypair,
```
thomaseizinger, line 158: Calling this `identity` would make it easier to reason about.

src/network.rs
```diff
@@ -151,6 +153,12 @@ pub struct Nectar {
...
+    #[behaviour(ignore)]
+    peer_id: PeerId,
```
thomaseizinger, line 160: Could compute this on demand, then you don't have to store it :)

thomaseizinger (approved):

rishflab (approved):

src/network.rs
D4nte: hehe, yes this was the name in one of my patches. 👍

src/network.rs
D4nte: I trust you on this one.

D4nte: Would it? Because there are quite a few `identity`s in this module (comit ones).

PR #116, opened 2020-09-09
Derive Bitcoin transient keys using BIP32

D4nte:
As per my [email on the mailing list](https://groups.google.com/a/comit.network/g/comit-dev/c/EUfgMFTU2jc/m/RF6YG99iAQAJ), I was not convinced about our usage of swap id to derive the bitcoin transient keys.

After more thought on the matter, adding a condition of knowing the swap id to be able to derive the bitcoin transient keys is adding one more barrier towards facilitating ongoing swap recoveries when the database is lost.

The other barrier is knowing the other party's Bitcoin address and the secret hash.

While this PR does not allow recovery of ongoing swaps after DB loss, it does make it slightly easier.
We also use a more standard way of generating keys (BIP32) than before.

Finally, one of the key change here is that the seed is not *always* store in memory in the `bitcoin::Wallet`. Which means a bit less changes of leaking or cold boot attacks, do not it does not prevent them, it's just moving towards some better security practices.

Some clean up of the network also done. Thanks @thomaseizinger for the fun and interesting pair programming session.


src/command/wallet_info.rs
```diff
@@ -2,6 +2,7 @@ use crate::{bitcoin, ethereum, Seed};
 
 pub async fn wallet_info(
     ethereum_wallet: Option<ethereum::Wallet>,
+    // TODO: have a "light wallet" here
```
thomaseizinger, line 5: Is this still relevant now?

src/swap/db.rs
```diff
@@ -62,6 +62,10 @@ impl Database {
 
         Ok(Database { db, tmp_dir })
     }
+
+    pub fn fetch_inc_bitcoin_transient_key_index(&self) -> anyhow::Result<u32> {
+        todo!()
```
thomaseizinger, line 67: Missing implementation :)

src/bitcoin/wallet.rs
```diff
@@ -1,46 +1,70 @@
...
+    fn fetch_inc(&self) -> anyhow::Result<u32>;
+}
+
+// TODO: cfg(test)
+impl TransientKeyIndexFetchInc for AtomicU32 {
+    fn fetch_inc(&self) -> anyhow::Result<u32> {
+        Ok(self.fetch_add(1, Ordering::SeqCst))
+    }
+}
+
```
thomaseizinger, line 41: Do we still need this now that you can just pass the index to the function?

thomaseizinger (commented):
Nice work, thanks Franck!

I believe you overlooked a few things though :)

D4nte:
Please note that to tackle the added `todo` to flush the db, I need to extract the business logic from the network behaviour to allow it to be async.
I'll do that in a follow-up PR.

README.md
```diff
@@ -3,3 +3,13 @@
 `nectar help` to see the available commands.
 
 Check `./sample-config.toml` if you want to customize the config.
+
+## Security advisory
+
+-   The seed file is used to generate the Bitcoin and Ethereum wallets.
+-   Bitcoin funds are held in a new wallet generated in the bitcoind instance. Keep your bitcoind instance secure.
```
tcharding, line 10: Two sentences on a single line, also there is whitespace at the end of this line.

README.md
```diff
@@ -3,3 +3,13 @@
 `nectar help` to see the available commands.
 
 Check `./sample-config.toml` if you want to customize the config.
+
+## Security advisory
+
+-   The seed file is used to generate the Bitcoin and Ethereum wallets.
+-   Bitcoin funds are held in a new wallet generated in the bitcoind instance. Keep your bitcoind instance secure. 
+-   If the seed file is lost, then any funds present in Nectar's Ethereum wallet is lost.
```
tcharding, line 11:
```suggestion
-   If the seed file is lost, then any funds present in Nectar's Ethereum wallet are lost.
```

README.md
```diff
@@ -3,3 +3,13 @@
...
+-   If the seed file is lost, Bitcoin funds can be recover from the bitcoind instance.
```
tcharding, line 12:
```suggestion
-   If the seed file is lost, Bitcoin funds can be recovered from the bitcoind instance.
```

README.md
```diff
@@ -3,3 +3,13 @@
...
+-   The bitcoind wallet is **not** password protected.
+-   If the database files are lost, then it is not possible to resume or abort ongoing swaps.
+-   If the database files are lost, once funds are recovered, a new seed should be generated to avoid re-using of the Bitcoin transient keys used in the HTLCs.
```
tcharding, line 15:
```suggestion
-   If the database files are lost, once funds are recovered, a new seed should be generated to avoid reuse of the Bitcoin transient keys used in the HTLCs. 
```

tcharding (approved):

rishflab (approved):

